PR #1, opened 2026-04-01
Add initial plan for Evidence Search Application

danielbreves:
Adds a plan for the new Evidence Search Application. The PRD outlines the architecture, core data model, and a detailed, step-by-step breakdown of the development tasks required to build a React-based frontend for searching and viewing ESEA-enhanced records from the destiny repository. It also specifies integration points, deployment requirements, and references to relevant backend and vocabulary artifacts.

### DIFF
--- a/docs/prd-evidence-search-application.md
+++ b/docs/prd-evidence-search-application.md
@@ -1,0 +1,249 @@
+# Epic: Evidence Search Application
+
+## Summary
+
+Build a new public-facing application for evidence synthesisers to search and inspect ESEA-enhanced records in the destiny repository. Static React app, deployed to blob storage with Azure Front Door. Auth via shared Keycloak infrastructure (same as taxonomy-builder) — needed for abstract visibility. Hard-coded for ESEA first.
+
+The app consumes the existing destiny-repository API for references/search and fetches vocabulary artifacts directly from blob storage for client-side label resolution (available as compact JSON-LD, TTL, and other formats). No BFF initially — introduce one later if client-side parsing proves insufficient.
+
+**Repo:** [destiny-evidence/evidence-repo-ui](https://github.com/destiny-evidence/evidence-repo-ui)
+
+## Architecture
+
+```
+┌──────────────────────────────────────────────────┐
+│  Static React App (Vite + TypeScript)            │
+│  Deployed: blob storage + Azure Front Door       │
+│  Auth: Keycloak (shared infra)                   │
+├──────────────────────────────────────────────────┤
+│  Consumes:                                       │
+│  1. destiny-repository API (search, references)  │
+│  2. vocabulary_uri → vocabulary artifact          │
+│  3. context_uri → JSON-LD context (immutable)    │
+│  Auth: Keycloak (shared infra)                   │
+└──────────────────────────────────────────────────┘
+```
+
+## EV Core class hierarchy (reference for card design)
+
+```
+Investigation
+  ├─ documentType, funderResearch, isRetracted, dataSource
+  └─ hasFinding → Finding[] (1+)
+       ├─ evaluates → Intervention
+       │   ├─ educationTheme[], implementerType, implementationDescription[]
+       │   ├─ implementationFidelity, implementationName, duration
+       │   └─ funderIntervention
+       ├─ comparedTo → ControlCondition
+       ├─ hasContext → Context (1+)
+       │   ├─ educationLevel[], setting[], country, countryLevel1
+       │   └─ participants[]
+       ├─ hasOutcome → Outcome (1+)
+       │   └─ outcome[] (CodingAnnotation)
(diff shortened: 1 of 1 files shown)


docs/prd-evidence-search-application.md
```diff
@@ -0,0 +1,247 @@
...
+
+**Repo:** evidence-repo-ui
+
+Scaffold the application:
+- Suggested stack: Vite + React + TypeScript (final choices left to implementer)
+- Client-side routing
+- HTTP client for API calls
+- Basic layout shell (header, main content area)
+- Environment config for destiny-repository API base URL and Keycloak
+- Auth integration via Keycloak (shared infrastructure, same as taxonomy-builder)
```
Adam-Hammo: Does it need authentication in the first iteration?

docs/prd-evidence-search-application.md
```diff
@@ -0,0 +1,247 @@
...
+
+---
+
+### evidence-repo-ui#2 — API client for destiny-repository
+
+**Repo:** evidence-repo-ui
+
+Create `src/services/apiClient.ts`:
+- `searchReferences(query, filters)` → calls `GET /v1/search/` with configurable base URL
+- `getReference(id)` → calls `GET /v1/references/{id}/`
```
Adam-Hammo: What's the direct reference lookup required for?

Adam-Hammo: Also a nit, I think `GET /v1/references/?identifier={id}` is better tested/supported as it does projecting etc - we should probably remove/hide the path-based one.

docs/prd-evidence-search-application.md
```diff
@@ -0,0 +1,247 @@
...
+---
+
+## Tasks
+
+### evidence-repo-ui#1 — Project scaffold and infrastructure
+
+**Repo:** evidence-repo-ui
+
+Scaffold the application:
+- Suggested stack: Vite + React + TypeScript (final choices left to implementer)
```
Adam-Hammo, line 61: Do we want a service worker to cache vocabs?

docs/prd-evidence-search-application.md
```diff
@@ -0,0 +1,247 @@
...
+**Repo:** evidence-repo-ui
+
+Create `src/services/investigationParser.ts`:
+
+Walks the `LinkedDataEnhancement.data` JSON-LD dict and produces typed structures using the vocabulary service for label resolution.
+
+**Input:** raw `data` dict from LinkedDataEnhancement + vocabulary service (using the enhancement's own `vocabulary_uri` — different enhancements may reference different vocabulary versions)
+**Output:** `InvestigationData` with typed `FindingData[]`, each containing `InterventionData`, `ContextData`, `OutcomeData`, `EffectEstimateData[]`, `ArmData[]`
+
+The graph structure is defined by the SHACL shapes (`evrepo-core-shapes.ttl`) and the `@context`. The parser should walk according to those. Implementation notes for non-obvious aspects:
```
Adam-Hammo, line 119: Should we expose `evrepo-core-shapes` somewhere? The amount of duplication is making me scared

docs/prd-evidence-search-application.md
danielbreves: Would it be worth the effort, considering they come from blob storage and are immutable?

docs/prd-evidence-search-application.md
danielbreves: My understanding is that we need auth to restrict access to abstracts at least. I guess there may be value in a public experience with unrestricted content, but that might be more work than providing access to everything behind auth for now?

docs/prd-evidence-search-application.md
Adam-Hammo: It would just save the network call, but you're right that's a very diminished return given the amount of API calling etc this UI will also need to do!

docs/prd-evidence-search-application.md
Adam-Hammo: Good point, let's lie on the safe side for now. To that point, should this be auth integration via Azure for now as the repository doesn't use Keycloak yet?

docs/prd-evidence-search-application.md
danielbreves: Deep-linking perhaps?

docs/prd-evidence-search-application.md
danielbreves: I've added a task to the plan.

docs/prd-evidence-search-application.md
danielbreves: Urgh I chose to forget about that. Would it be too optimistic to make the switch over blocking? 😰

Adam-Hammo: Probably one to discuss at tomorrow's tech meet - we gotta do it someday! I'd definitely be in favour of doing it "soon" but prioritising it over this UI is probably a hard sell.

Issue #7, opened 2026-04-08
Search page and results list


danielbreves:
Implement the search page at `/:community`. Only `/esea` will be functional initially.

### Search

- Search box with query input
- Scoped to the community in the URL path — filters to that community's enhanced records
- Concept filtering via Lucene query syntax against projected fields; dedicated filter parameters deferred

### Results list

- Terse list: title, authors, year, DOI link, abstract snippet
- Pagination
- Possible richness indicators (finding count, whether effect estimates are present) — TBD, take a rough shot and iterate
- Click result → navigate to `/:community/references/:id`

mootpointer:
## Design Spec

**Prototype:** [Search page](https://taxonomy-prototype.pages.dev/demo/search) · [Investigation detail](https://taxonomy-prototype.pages.dev/demo/investigation)

> **Note:** The hosted prototype uses the previous font stack (Newsreader/Manrope/JetBrains Mono) and warm color scheme. The current design direction uses the Klim fonts and cool palette described below. See `demo/search-klim-b.html` locally for the latest cut.

### Layout

- Sticky app header with logo mark, community name ("Evidence Repository / Education"), and auth placeholder
- Search hero: centered Signifier title, subtitle in italic, search bar with year range filter
- Results sit inside a white card container (`#FFFFFF`) on the cool gray background (`#F4F5F7`)
- Alternating row tint (`#FAFBFC`) on even rows for eye tracking
- Hover state: `#EAECEF`

### Result row structure

Each row is a grid: content on the left, metadata on the right.

**Left column:**
1. Title (Signifier, 17px, weight 400) with year in muted parenthetical
2. Authors (Untitled Sans 13px, secondary color)
3. Venue (13px, italic, tertiary color)
4. Abstract (13px, secondary color, 2-line clamp via `-webkit-line-clamp: 2`)
5. Concept tags — ghost outlined style (see below)

**Right column (flex column, right-aligned):**
1. DOI link (Söhne Mono 11px, fades in on row hover)
2. Finding count badge
3. Effect estimate count badge

### Tags on results

Tags use a single neutral ghost treatment — **not** filled color pastels:
- Background: transparent
- Border: 1px solid `#B0B8C4`
- Text: `#424A53`
- Border-radius: 3px
- Font: 12px, weight 500

All tags look the same regardless of property type. Color differentiation is not used.

### Pagination

[...]

PR #13, opened 2026-04-09
API client for destiny-repository

### LINKED ISSUE #4: API client for destiny-repository
Create `src/services/apiClient.ts`:

- `searchReferences(query, filters)` → calls `GET /v1/search/` with configurable base URL
- `getReference(id)` → calls `GET /v1/references/{id}/`
- Helper to extract `LinkedDataEnhancement` and `BibliographicMetadataEnhancement` from the enhancements array (use most recent if multiple)
- TypeScript types for API response shapes (Reference, Enhancement, SearchResult, pagination)

danielbreves:
## Summary
- Expand `src/types/models.ts` with full TypeScript types matching the backend SDK (Reference, Enhancement, SearchResult, pagination, etc.)
- Add `src/services/apiClient.ts` with `searchReferences()`, `getReference()`, and helpers to extract the most recent bibliographic/linked-data enhancements
- Add unit tests covering all API client functions and extraction helpers (15 tests)

Closes #4

## Test plan
- [x] `npm run typecheck` passes
- [x] `npm run test` — all 16 tests pass (15 new + 1 existing)

### DIFF
--- a/src/services/apiClient.ts
+++ b/src/services/apiClient.ts
@@ -1,0 +1,64 @@
+import { api } from "@/api/client";
+import type {
+  Reference,
+  SearchResult,
+  Enhancement,
+  BibliographicMetadataEnhancement,
+  LinkedDataEnhancement,
+} from "@/types/models";
+
+export interface SearchFilters {
+  page?: number;
+  annotation?: string[];
+  sort?: string[];
+}
+
+export async function searchReferences(
+  query: string,
+  filters: SearchFilters = {},
+): Promise<SearchResult> {
+  const params = new URLSearchParams();
+  params.set("q", query);
+  if (filters.page) {
+    params.set("page", String(filters.page));
+  }
+  if (filters.annotation) {
+    for (const a of filters.annotation) {
+      params.append("annotation", a);
+    }
+  }
+  if (filters.sort) {
+    for (const s of filters.sort) {
+      params.append("sort", s);
+    }
+  }
+  return api.get<SearchResult>(`/v1/search/?${params.toString()}`);
+}
+
+export async function getReference(id: string): Promise<Reference> {
+  return api.get<Reference>(`/v1/references/${id}/`);
+}
+
+export function extractBibliographic(
+  reference: Reference,
+): BibliographicMetadataEnhancement | null {
+  if (!reference.enhancements) return null;
+  const matches = reference.enhancements.filter(
+    (e): e is Enhancement & { content: BibliographicMetadataEnhancement } =>
+      e.content.enhancement_type === "bibliographic",
+  );
+  if (matches.length === 0) return null;
+  return matches[matches.length - 1].content;
+}
+
+export function extractLinkedData(
+  reference: Reference,
+): LinkedDataEnhancement | null {
+  if (!reference.enhancements) return null;
+  const matches = reference.enhancements.filter(
+    (e): e is Enhancement & { content: LinkedDataEnhancement } =>
+      e.content.enhancement_type === "linked_data",
+  );
+  if (matches.length === 0) return null;
+  return matches[matches.length - 1].content;
+}
--- a/src/services/apiClient.ts
+++ b/src/services/apiClient.ts
@@ -3,6 +3,7 @@ import type {
   Reference,
   SearchResult,
   Enhancement,
+  EnhancementContent,
   BibliographicMetadataEnhancement,
   LinkedDataEnhancement,
 } from "@/types/models";
@@ -19,7 +20,7 @@ export async function searchReferences(
 ): Promise<SearchResult> {
   const params = new URLSearchParams();
   params.set("q", query);
-  if (filters.page) {
+  if (filters.page !== undefined) {
     params.set("page", String(filters.page));
   }
   if (filters.annotation) {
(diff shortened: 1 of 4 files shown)


src/services/apiClient.ts
```diff
@@ -0,0 +1,64 @@
...
+    for (const a of filters.annotation) {
+      params.append("annotation", a);
+    }
+  }
+  if (filters.sort) {
+    for (const s of filters.sort) {
+      params.append("sort", s);
+    }
+  }
+  return api.get<SearchResult>(`/v1/search/?${params.toString()}`);
```
jaybea: I believe the search endpoint is /v1/references/search, not /v1/search. 

I Verified against app/domain/references/routes.py in destiny-repository — the search_router (prefix /search) is included into reference_router (prefix /references), so the assembled path is /v1/references/search/.

(Unless I'm totally lost!)

src/services/apiClient.ts
```diff
@@ -0,0 +1,64 @@
...
+export function extractBibliographic(
+  reference: Reference,
+): BibliographicMetadataEnhancement | null {
+  if (!reference.enhancements) return null;
+  const matches = reference.enhancements.filter(
+    (e): e is Enhancement & { content: BibliographicMetadataEnhancement } =>
+      e.content.enhancement_type === "bibliographic",
+  );
+  if (matches.length === 0) return null;
+  return matches[matches.length - 1].content;
```
jaybea: Use the most recent if multiple" is implemented as "take the last element," but the destiny-repository API doesn't guarantee enhancement ordering. I checked app/domain/references/models/sql.py:84 — the SQLAlchemy relationship has no ORDER BY, the API route doesn't sort, and deduplicated references concatenate enhancements from multiple sources without ordering.

Just recently, each Enhancement has a created_at timestamp (visible in the API response, e.g., "created_at":"2026-04-12T01:01:08.922174Z") - it just isn't included in the Enhancement TypeScript type yet.

```typescript
const match = reference.enhancements
  .filter((e): e is Enhancement & { content: T } => e.content.enhancement_type === enhancementType)
  .sort((a, b) => a.created_at.localeCompare(b.created_at))
  .at(-1);
```

src/services/apiClient.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export function extractLinkedData(
```
jaybea, line 71: Could a generic extractEnhancement<T>(ref, type) collapse these two (extractBibliographic and extractLinkedData)?

src/services/apiClient.ts
```diff
@@ -0,0 +1,64 @@
...
+  sort?: string[];
+}
+
+export async function searchReferences(
+  query: string,
+  filters: SearchFilters = {},
+): Promise<SearchResult> {
+  const params = new URLSearchParams();
+  params.set("q", query);
+  if (filters.page) {
```
jaybea: `if (filters.page)` drops `page: 0`. If the API ever uses 0-based pagination (or you want to allow an explicit "first page" via `page: 0`), this silently fails.

jaybea (changes requested):
Nice clean PR. The types are well-structured, the tests cover the right edge cases, and the layering on top of `api/client.ts` is clean. I pulled the branch down and ran it against a local destiny-repository instance to verify the API contract end-to-end. Most things work as expected, but there are a few issues worth addressing before merge.

**What I verified against real destiny-repository:**
- `searchReferences` returned 87 results for `q=phonics`, with the expected `total`/`page`/`references` shape
- `searchReferences` returned 2,392 ESEA-coded references for `q=linked_data_concepts:*C00008*`
- `getReference` fetches individual references by UUID
- `extractBibliographic` returned real bibliographic data (e.g., "Effects of the Active Smarter Kids (ASK)..." 2019)
- `extractLinkedData` returned full investigation graphs with 4 findings, effect estimates with Hedges' g and CIs, and the JSON-LD structure described in the PRD

**Issues caught by running the code:**
1. Search endpoint path is wrong — `/v1/search/` should be `/v1/references/search/` (the unit tests pass because they mock `api.get`, but the real call 404s)
2. "Most recent enhancement" relies on array position, but the API doesn't guarantee order — the `Enhancement` payload includes `created_at` which can be used to sort

**One question, not a request for change** — `src/App.tsx:9-11`, `SearchPage.tsx`, `RecordDetailPage.tsx`:

The scaffold branch had a guardrail chain — `findCommunity()` lookup against a static whitelist, `<NotFoundPage default />` route, and four route tests. This PR removes all of it. As of this branch, `http://localhost:3000/banana` renders "Search banana" with no error, and unmatched routes (including `/`) render an empty shell.

Was this a deliberate clean slate (the validation will come back in a different form when search lands), or did this slip in by accident? Either is reasonable, just wanted to confirm the intent before merge so we don't lose the behavior unnoticed.

**Out of scope but flagging:** on `origin/api-client`, `vite.config.ts:18` defaults the proxy target to `http://localhost:8080`, but local DR runs on 8000. There's also no `rewrite` rule, so `/api/v1/...` is forwarded as-is and 404s against DR (which has no `/api` prefix). Both are pre-existing on the scaffold branch — not this PR's regression — but they mean the dev server can't hit a local DR out of the box. Worth a separate small PR?

src/services/apiClient.ts
```diff
@@ -0,0 +1,64 @@
...
+  if (filters.sort) {
+    for (const s of filters.sort) {
+      params.append("sort", s);
+    }
+  }
+  return api.get<SearchResult>(`/v1/search/?${params.toString()}`);
+}
+
+export async function getReference(id: string): Promise<Reference> {
+  return api.get<Reference>(`/v1/references/${id}/`);
```
Adam-Hammo: I think we should consider using [`/v1/references/?identifier=${id}`](https://github.com/destiny-evidence/destiny-repository/tree/main/libs/ad_hoc/scripts/deduplicate_eef_references) (even though it's marked as experimental) - this endpoint returns the deduplicated projection, unlike the direct endpoint which just returns a naked `Reference`.

src/services/apiClient.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export interface SearchFilters {
+  page?: number;
+  annotation?: string[];
+  sort?: string[];
+}
+
+export async function searchReferences(
+  query: string,
+  filters: SearchFilters = {},
```
Adam-Hammo, line 19: Annoying architectural question - when it comes to different communities, I imagine they will have different "default" annotation filters to search down to their domain. Should that live here? 

One implementation might look like importing the `community` enum, mapping it to the required annotation filter, and moving the API methods to a class that takes in the community on instantiation automatically attaches the filter. But there's certainly many other ways we could do it!

danielbreves: That makes, but perhaps we can address it in https://github.com/destiny-evidence/evidence-repo-ui/issues/7 when we implement search? Happy to copy your comment there if you agree.

danielbreves:
> The scaffold branch had a guardrail chain — findCommunity() lookup against a static whitelist, <NotFoundPage default /> route, and four route tests. This PR removes all of it.

@jaybea these are still here, no? https://github.com/destiny-evidence/evidence-repo-ui/blob/71746dde1f49b4208b4e0e6b6b7ee002d17ca9d3/src/pages/SearchPage.tsx#L11-L12

https://github.com/destiny-evidence/evidence-repo-ui/blob/71746dde1f49b4208b4e0e6b6b7ee002d17ca9d3/tests/App.test.tsx#L21-L25

jaybea (approved):
Yes, they're there. I reviewed an out-of-date version of the branch previously, the pre-rebase commit a25c181. That one was missing the community validation files because they’d been added to scaffold after your branch point. I've looked at the current commit now, everything’s in place.

I re-verified against live DR: search returns 87 results for q=phonics, the identifier endpoint returns the deduplicated projection, and created_at sort handles the real-world dedup case properly. I saw one reference with two sets of bibliographic/abstract/raw enhancements from merged duplicates, about 58ms apart.

Thanks for picking up the Vite proxy fix, That was outside the original review scope and saves a separate PR. And the swap to `/v1/references/?identifier=` (per Adam's suggestion) is looks like the right way to go. Gives us the deduplicated projection rather than the naked reference, which pairs well with the `created_at` sort for handling merged duplicates cleanly.

Looks good to me. Approving.

PR #14, opened 2026-04-09
Add deployment pipeline and Terraform infrastructure

danielbreves:
## Summary

- GitHub Actions workflows: `tests.yml` (CI on PRs/main), `deploy.yml` (build + deploy to Azure Blob Storage + Front Door cache purge), `promote-to-production.yml` (manual production promotion)
- Terraform infrastructure (`infra/app/`): storage account with static website hosting, Front Door endpoint with cache rules (immutable hashed assets, no-cache HTML), OIDC federated credentials for GitHub Actions, DNS via DNSimple, and GitHub environment variables provisioned automatically

Follows the taxonomy-builder deployment pattern adapted for a standalone static SPA (no backend, containers, or database).

## Test plan

- [x] Create Terraform Cloud workspace with `evidence-repo-ui` tag
- [x] `terraform plan` against a target environment to verify resource configuration
- [x] Push a branch / open PR to verify `tests.yml` runs
- [ ] Merge to main to verify `deploy.yml` triggers staging deployment
- [ ] Manual dispatch `deploy.yml` to verify deployment
- [ ] Verify deployed site loads at the Front Door URL


.github/workflows/promote-to-production.yml
```diff
@@ -0,0 +1,17 @@
...
+on: workflow_dispatch
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  deploy:
+    uses: ./.github/workflows/deploy.yml
+    with:
```
jack-o-the-green: This will currently rebuild instead of promoting pre-existing build artifacts (there's no restrictions on the build job in deploy.yml`. Fine if we want this, but if we do we should just ditch this file and use the deploy action.

infra/app/locals.tf
```diff
@@ -0,0 +1,15 @@
+locals {
+  name          = "${var.app_name}-${var.environment}"
+  name_short    = "${replace(var.app_name, "-", "")}${substr(var.environment, 0, 4)}"
+  environment_short = substr(var.environment, 0, 4)
+  is_production     = var.environment == "production"
+  custom_domain     = local.is_production ? "${var.subdomain}.${var.custom_domain}" : "${var.subdomain}.${local.environment_short}.${var.custom_domain}"
+
+  minimum_resource_tags = {
```
jack-o-the-green: These tags were initially required for UCL deployments but are now propagating their way through our repos. Not sure they provide a lot of value and we need to add a bunch of extra variables for them. Happy to leave but could also remove!

infra/app/main.tf
```diff
@@ -0,0 +1,6 @@
+# Core resource group for all evidence-repo-ui resources
+resource "azurerm_resource_group" "this" {
+  name     = "rg-${local.name}"
+  location = var.region
+  tags     = merge({ "Budget Code" = var.budget_code }, local.minimum_resource_tags)
```
jack-o-the-green: Same as above re tags

infra/app/storage.tf
```diff
@@ -0,0 +1,30 @@
+# Storage account for static website hosting
+resource "azurerm_storage_account" "frontend" {
+  # Storage account name must be 3-24 chars, lowercase letters and numbers only
+  name                     = "${local.name_short}fe"
```
jack-o-the-green: `nitpick` most resources have the resource type prefix at the beginning i.e. this would be `fe${local.name_short}` instead

infra/app/terraform.tf
```diff
@@ -0,0 +1,55 @@
+terraform {
+  required_version = ">= 1.0"
```
jack-o-the-green: Feel like we should add a minor version in here, most of us should be using `>=1.14.0` by now

jack-o-the-green (commented):
Mostly nitpicks, looks reasonable to me!

infra/app/locals.tf
danielbreves: So they're not required anymore?

jack-o-the-green: We're not deployed in the UCL tenant, so they're not required. We'd need to add them if we moved over but I think that's something to handle when and if it happens. Otherwise it's just more variables.

infra/app/frontdoor.tf
```diff
@@ -0,0 +1,182 @@
...
+}
+
+# Origin
+resource "azurerm_cdn_frontdoor_origin" "frontend" {
+  name                          = "o-frontend-${local.name}"
+  cdn_frontdoor_origin_group_id = azurerm_cdn_frontdoor_origin_group.frontend.id
+
+  enabled                        = true
+  host_name                      = azurerm_storage_account.frontend.primary_web_host
+  http_port                      = 80
```
mootpointer: Nitpick, but do we need to support the HTTP port?

infra/app/frontdoor.tf
```diff
@@ -0,0 +1,182 @@
...
+  patterns_to_match      = ["/*"]
+  forwarding_protocol    = "HttpsOnly"
+  link_to_default_domain = true
+  https_redirect_enabled = true
+
+  cdn_frontdoor_custom_domain_ids = [azurerm_cdn_frontdoor_custom_domain.this.id]
+  cdn_frontdoor_rule_set_ids      = [azurerm_cdn_frontdoor_rule_set.cache.id]
+
+  cache {
+    query_string_caching_behavior = "IgnoreQueryString"
```
mootpointer: Is keeping query strings useful to allow us to do our own cache busting if it's necessary?

mootpointer (approved):
Looks good. Just some really minor things to *maybe* tweak

infra/app/frontdoor.tf
danielbreves: We have `https_redirect_enabled` below, this is just setting the value of the ports to the default, so it's effectively no-op. Happy to remove.

PR #15, opened 2026-04-14
Vocabulary service (client-side label resolution)

### LINKED ISSUE #5: Vocabulary service (client-side label resolution)
Create a vocabulary service for client-side label resolution.

### Vocabulary fetching + caching

- Fetch vocabulary from `vocabulary_uri` (HTTP GET, immutable URL — cache indefinitely)
- The published vocabulary is available in multiple formats: compact JSON-LD (`vocabulary.jsonld` with `@context` + `@graph`), TTL (`vocabulary.ttl`), plain JSON (`vocabulary.json`), RDF/XML (`vocabulary.rdf`)
- Build `Map<conceptUri, prefLabel>` from SKOS Concept entries
- Approach options: parse compact JSON-LD directly, or use rdflib.js with TTL. Try the simplest approach first (JSON-LD compact is parseable as plain JSON); fall back to rdflib.js if needed
- evrepo-core concepts (effect size metrics, estimate sources) are imported into the ESEA vocabulary, so no separate static mapping is needed

### Context fetching + URI expansion

- Fetch `@context` JSON from the context URI in the linked data (`context.jsonld`)
- Cache by URL
- Implement `expandUri(compactUri)` using prefix mappings from the context (e.g. `esea:C00008` → `https://vocab.esea.education/C00008`)

**Reference:** Published artifacts at `taxonomy-builder/backend/.blob-storage/019d2cfa-.../1.0/` (vocabulary.jsonld, vocabulary.ttl, context.jsonld). ESEA context is at `destiny-repository/app/static/vocab/esea/esea-context.jsonld`.

danielbreves:
## Summary

- Add vocabulary service that fetches and parses vocabulary.jsonld as plain JSON, building a concept URI → prefLabel map
- Add context service that fetches JSON-LD context documents, extracts prefix mappings, and provides `expandCompactUri()` for resolving compact URIs to full URIs
- Add `useVocabulary` Preact hook for component consumption
- Zero additional runtime dependencies

Closes #5

## Test plan

- [x] `npm test` — 17 tests pass (vocabularyService, contextService, useVocabulary hook)
- [x] Bundle size: no impact (vocabulary service is not yet imported from entry point)
- [ ] End-to-end: verify label resolution works with published ESEA vocabulary once integrated into record detail page


jaybea (approved):
Nice clean PR. The split between vocabulary and context services is tidy, the cache eviction on rejection is correct, and dropping `rdflib` for plain JSON parsing keeps the bundle light.

**I verified this e2e against my local destiny-repository + vocabulary publisher:**

* `fetchVocabulary` against `http://localhost:3001/.../vocabulary.ttl` → normalises to `.jsonld`, fetches, parses, returns **181 ESEA `skos:Concept` labels** (for example `https://vocab.esea.education/SampleFeaturesScheme/C00191` → "Learning Differences")
* Same flow against `https://vocab.staging.evidence-repository.org/.../vocabulary.ttl` → 200 OK
* `fetchContext` extracts **47 prefix mappings** from `context.jsonld`
* `expandCompactUri("esea:DocumentTypeScheme/C00008", prefixes)` → `https://vocab.esea.education/DocumentTypeScheme/C00008` ✅
* `npm run typecheck` clean, 38/38 tests pass

Approving. Two small notes below, both pretty take-it-or-leave-it.

---

**Note 1 (low) - `vocabularyService.ts:17` URL regex is a bit broader than intended.**

```ts
url.pathname.replace(/\.\w+$/, "")
```

This strips any trailing `.word` from the pathname, not just known serialisation extensions. It works fine on every URL DR currently emits, because they all end in `vocabulary.{ext}`. But if an LDE ever carried a URL like:

* `.../1.0` → becomes `.../1.jsonld` (strips `.0`)
* `.../releases/2026.04` → becomes `.../releases/2026.jsonld` (strips `.04`)

Tightening it to known extensions would harden this a little against future LDE shape changes:

```ts
url.pathname.replace(/\.(jsonld|json|ttl|rdf|xml)$/, "")
```

Not a current bug, just being a bit defensive.

---

**Note 2 (medium) - `useVocabulary.ts:15-24` and `useContextPrefixes.ts:16-23` have a subtle race on URL change.**

The comment is right that vocabulary URLs are immutable in content, but the prop reference can still change at the component level. For example, if the user navigates between two references that point at different vocabulary versions. With no cancellation guard. I think the unavoidable reality is that we will have LDE's against different vocabularies eventually:

1. Render with URL A → fetch A starts
2. Re-render with URL B → fetch B starts
3. Fetch A resolves later → `setLabels(A)` overwrites B's labels

Also, when the URL transitions to `undefined`, the previous `labels` value sticks around rather than clearing.

I think the usual fix is just the standard cancellation flag, if you want to deal with it now. Also totally fair to defer if real navigation between vocabulary versions is going to be very rare for a while.

```ts
useEffect(() => {
  if (!vocabularyUrl) { setLabels(null); return; }
  let cancelled = false;
  setLoading(true);
  setError(null);
  getCachedVocabulary(vocabularyUrl)
    .then(r => { if (!cancelled) setLabels(r); })
    .catch(e => { if (!cancelled) setError(e); })
    .finally(() => { if (!cancelled) setLoading(false); });
  return () => { cancelled = true; };
}, [vocabularyUrl]);
```

PR #18, opened 2026-04-15
Investigation parser and card

### LINKED ISSUE #8: Record detail page: Investigation parser + card
Implement the investigation parser and investigation card on `/:community/references/:id`.

## Investigation data parser

Create `src/services/investigationParser.ts`:

Walks the `LinkedDataEnhancement.data` JSON-LD dict and produces typed structures using the vocabulary service for label resolution.

**Input:** raw `data` dict from LinkedDataEnhancement + vocabulary service (using the enhancement's own `vocabulary_uri` — different enhancements may reference different vocabulary versions)
**Output:** `InvestigationData` with typed `FindingData[]`, each containing `InterventionData`, `ContextData`, `OutcomeData`, `EffectEstimateData[]`, `ArmData[]`

### Implementation notes

- Compact URIs (e.g. `esea:C00008`) need expanding via the `@context` before vocabulary lookup
- `NumericCodingAnnotation` uses typed JSON-LD values (`{"@type": "xsd:integer", "@value": 47}`), not bare numbers
- Status values are compact URIs too (`"evrepo:coded"`, `"evrepo:notReported"`)
- Blank node IDs (`"_:intervention"`, `"_:control"`) cross-reference arms to conditions — collect on first pass, resolve on second
- Skip annotations with status `notReported` or `notApplicable`

**Reference data:**
- Robot output examples: `vocabulary-mapping-robot/vocab/0e8f0ddd-d0ab-460c-ac66-2683a11f3ef7.jsonld`, `vocabulary-mapping-robot/vocab/7d6f1092-1976-4903-ab17-330b79787f7d.jsonld`, `vocabulary-mapping-robot/vocab/b7a34d6d-f11e-4852-9030-d060a6bfdf8f.jsonld`
- SHACL shapes: `destiny-repository/app/static/evrepo-core-shapes.ttl`

## Investigation card

- Fetch reference, vocabulary, context (with caching)
- Parse linked data via investigation parser

### Investigation card renders

- Title, authors (formatted citation), publication venue, volume/issue/pages
- Publication year
- Tags row: documentType label, education level labels as concept pills
- Sample size, analysis sample from finding-level data
- Funder section: funderResearch, funderIntervention
- isRetracted flag (prominent warning if true)

Concept tags show property label + concept label (e.g. "education level: Lower Primary"), linkable to taxonomy definition URLs.

danielbreves:
## Summary

- Add investigation parser that walks `LinkedDataEnhancement` JSON-LD, resolves blank nodes across findings, expands compact URIs via fetched context, and resolves labels via vocabulary service
- Add investigation card component (title, citation-formatted authors, venue, DOI, document type tag, retracted banner)
- Add context service for fetching/caching JSON-LD context documents and prefix expansion
- Add reusable `TagGroup` component for labelled ghost-tag pattern
- Consolidate CSS design tokens with consistent naming
- Wire up `RecordDetailPage` with data fetching, parsing, and rendering
- Dev proxy for staging vocabulary artifacts (workaround until CORS enabled on taxonomy-builder, see destiny-evidence/taxonomy-builder#201)

Screenshot 2026-04-15 at 1 39 49 PM

## Test plan

- [x] `npm test` — 64 tests pass (investigation parser, context service, components, hooks)
- [x] Local dev: investigation card renders with title, authors, DOI, document type from staging vocabulary
- [ ] Verify with additional seed data (multiple findings, retracted state)

Closes #8


jaybea (changes requested):
Nice work on this. Sorry for the long review.

The record detail page is looking good. I especially like how the Investigation card is shaping up, and the context/vocabulary integration looks to be wired correctly from what I could tell. I ran `npm run typecheck` here, which was clean, and all 76 tests are passing.

I also poked at it in-browser against local destiny-repository on port 8000, with the dev server pointed at the published production vocabulary through the Vite dev proxy.

**What worked for me in the browser:**

* `019d986a-cb4a-76d4-ae12-114040e8e837` — card renders title, authors, year, DOI, and "DOC TYPE · Journal Article"
* `019d986a-8304-714b-91a7-fe187a63b57a` — multi-bib dedup reference (2 bibliographic enhancements from deduplicated ingestions with different title casing) renders correctly
* `019d9869-e3d5-…` and `019d9869-ed0a-…` — multi-bib dedup with 1 LDE each, render cleanly with doc-type tag
* DOI link resolves to the Frontiers article — `extractDoi` wiring looks right
* Vocab label resolution chain works end-to-end: `esea:DocumentTypeScheme/C00008` → expanded via context → resolved to "Journal Article" via vocabulary
* Bogus-but-well-formed UUID (`01944444-…`) → clean NotFoundPage (404 path looks right)
* Retracted fixture happy path (`019d9868-f418-…`, `isRetracted: true` on the Investigation node): renders the banner with `role="alert"` and `investigation-card--retracted` modifier class

**A few things that looked off to me:**

* `019d9869-2718-…` (abstract + bibliographic, no LDE, no DOI) — card renders title/authors/year but still shows the "Investigation" kicker and a dangling `<hr>` divider with nothing below it
* All-zero UUID (`00000000-0000-0000-0000-000000000000`) — raw backend error leaks to UI: *"Failed to load reference: Invalid identifier lookup string: … Must be UUID if no identifier type is specified."*
* SPA nav between two refs (via `history.pushState`) — I captured intermediate renders at t+30/80/150ms, and the prior ref's title and doc-type tag stay visible for a short moment while the new ref loads
* Retracted fixture with vocab fetch blocked in-browser (`/vocab-proxy/*` rejected in `window.fetch`, as a local simulation of vocab/context resolution failure) — same retracted URL renders with no banner, no retracted class, no doc-type tag, no loading state, and no error

A few findings below. Some of these may be deliberate tradeoffs, but they looked worth flagging.

---

**High: I think `referenceUtils.ts` lost the `created_at` sort from the extraction helpers.**

The refactor that moved `extractBibliographic` / `extractLinkedData` out of `apiClient.ts` into `referenceUtils.ts` seems to have dropped the ordering semantics. The originals in `apiClient.ts` sort by `created_at` to select the most recent enhancement, whereas the new copies in `referenceUtils.ts:9-31` just take the last array element, which I don’t think is guaranteed to be the newest.

`RecordDetailPage` is importing the new versions, so on references with multiple enhancements of the same type, this looks like it could surface stale bibliographic metadata or an older linked-data enhancement. We’ve already confirmed in real destiny-repository data that this case exists, since deduplicated references can carry merged enhancements from multiple ingestions.

Maybe I’m missing some upstream guarantee on ordering, but if not, I think this is a real regression.

The safest fix is probably to move the existing `extractEnhancement` implementation out of `apiClient.ts` into `referenceUtils.ts`, preserving the `created_at` sort and its tests, then have both callers import the same helper.

---

**High: vocabulary/context resolution failure currently looks like missing investigation data, which can hide retraction state.**

This one made me a bit nervous, although admittedly some things have to come together to make it happen, but hey, it's the internet, right?

As best I can tell, the causal chain is:

* `RecordDetailPage.tsx:29,34` ignores `error` from `useVocabulary` / `useContextPrefixes`
* `parseInvestigation` at line 37 is gated on successful label/context resolution (`context && labels`)
* so when either fetch fails, `investigation` becomes `null`
* and `isRetracted` falls back to `false` at line 78

That means anything investigation-derived can silently disappear: the retraction banner, the doc-type tag, and the `investigation-card--retracted` class modifier.

I verified this with the retracted fixture `019d9868-f418-…` where `isRetracted: true` is present on the Investigation node. On the happy path it renders correctly. When I block the vocab fetch in-browser, the same record renders with no banner, no retracted modifier class, no doc-type tag, and no visible error. In other words, a retracted investigation ends up looking like a normal one.

In an evidence-synthesis app, that feels a bit beyond ordinary degraded UX and starts to edge into correctness/safety territory.

Two fixes seem worth considering:

* **Parse raw investigation flags independently of label resolution.** `isRetracted` is just a boolean on the Investigation node, so it shouldn’t need vocabulary or context. If that gets parsed directly from `linkedData.data` as soon as the LDE is available, then a vocab publisher hiccup can’t hide the retraction state.
* **Surface `vocabError` / `ctxError` through `RecordDetailPage`.** Even a small visible "vocabulary unavailable" state would be better than silently dropping investigation-derived UI.

I may be overreacting a little, but this one felt important enough to call out as High.

---

**Medium: I think `useReference` should clear stale state when `id` changes or becomes `undefined`.**

The hook has a cancellation guard, which is good, but I think stale state is still visible during transitions.

I checked this in the browser by navigating from `019d9869-e3d5` → `019d9869-2718`, and for a short window the previous reference title and doc-type tag remain on screen before the new one appears. It is brief, but definitely noticeable if you watch for it.

During that window, `extractLinkedData(reference)` is still running against the previous reference, so it can feed stale `vocabulary_uri` and `contextUrl` into the downstream hooks. Also, if `id` becomes `undefined`, state doesn’t seem to get reset because of the early return at `useReference.ts:15`.

My guess is the simplest fix is to clear `reference` / `error` inside the effect, and also on the `!id` path, so consumers get a clean loading state on transition. Something roughly like:

```ts
useEffect(() => {
  if (!id) {
    setReference(null);
    setError(null);
    setLoading(false);
    return;
  }

  let cancelled = false;
  setReference(null);
  setError(null);
  setLoading(true);
  // ...existing fetch logic
}, [id]);
```

---

**Low: no-LDE records still show investigation framing/divider even when there’s no investigation content.**

This showed up in-browser on `019d9869-2718`.

`InvestigationCard.tsx:76` always renders the "Investigation" kicker, and line 98 always renders the divider. So for references that have no `linked_data` enhancement, the page ends up showing investigation framing with nothing investigation-ish underneath it.

That felt a little odd in use. When there’s no LDE, the page is really just showing bibliographic reference metadata, not an investigation.

A couple possible directions:

* hide the kicker + divider when there’s no investigation content, or
* separate the bibliographic header from the investigation section structurally, so the investigation section can disappear cleanly

Not a big deal, just looked a bit awkward.

---

**Low: raw backend error message is surfaced directly in the UI for malformed identifiers.**

Reproduces for any identifier that destiny-repository can't parse. Checking the endpoint directly, it rejects anything that isn't a valid UUIDv4 or UUIDv7 when `identifier_type` isn't specified — so the nil UUID, anything with an invalid version nibble (e.g. `01944444-4444-1444-…`), and obvious non-UUIDs like `abc` all hit this path. A valid-format v4/v7 UUID that simply doesn't exist in the DB returns an empty list and falls through to `NotFoundPage`, which is the correct behaviour.

As a concrete example, navigating to the app route `/esea/references/00000000-0000-0000-0000-000000000000` in the dev server displays:

> Failed to load reference: Invalid identifier lookup string: 00000000-0000-0000-0000-000000000000. Must be UUID if no identifier type is specified.

`RecordDetailPage.tsx:53-61` is rendering `refError.message` verbatim from destiny-repository. The 404 case falls through to `NotFoundPage` (via `getReference` returning `null`), so this is specifically the non-404 API error path.

I’d lean toward replacing that with something generic like "We couldn't load this reference" and maybe a retry/back link, while keeping `NotFoundPage` for the actual not-found case.

---

**Low: `TagGroup.tsx:18` uses `key={i}` — I think this used to be `key={tag}`.**

Not an issue today for the current single-tag usage, but index keys are usually a lurking gotcha! if the list ever becomes dynamic or reordered.

The previous `key={tag}` is better if tags are unique within a group. If duplicates are possible later, the caller probably needs stable tag IDs, or the component should de-dupe before rendering; ``key={`${tag}-${i}`}`` is only a fallback (unique per render, but not reorder-stable since the index is part of the key).

---

**Low: `investigationParser.ts:20` uses `split("/").pop()`, which won’t handle hash-style IRIs.**

Quick background (I had to look it up!) - RDF vocabularies use IRIs (Internationalized Resource Identifiers, basically URIs with Unicode) to identify concepts, and they come in two flavours:

* **Slash IRIs** - the local name is the last path segment. Our ESEA vocab currently does this:
  `https://vocab.evidence-repository.org/notReported` → local name `notReported`
* **Hash IRIs** - the local name is after a `#` fragment. Most W3C standards do this (OWL, RDFS, SKOS, etc.):
  `http://www.w3.org/2002/07/owl#Class` → local name `Class`

Both are legitimate conventions and it's common to end up with a mix once you start importing external vocabularies. The parser at `investigationParser.ts:20` uses `split("/").pop()` to extract the local name, which works for ESEA today but would return `www.w3.org#notReported` (the whole last path segment) if a hash IRI ever showed up — and the `SKIP_STATUSES` comparison would silently miss, so retracted/not-applicable statuses could end up rendering as if they were real values.

Not an immediate issue since we control the vocabulary, but `split(/[/#]/).pop()` covers both conventions in one line and is the kind of thing that's cheap now and annoying to debug (if missing) later.

---

**Note, but not new in this PR:** `useVocabulary` / `useContextPrefixes` still don’t have a cancellation guard on URL change. I think that was already noted on PR #15, so I’m not counting it as a new regression here.

danielbreves:
Thanks for the comprehensive review @jaybea! All addressed I think.

jaybea (approved):
Approving, and thanks for the quick turnaround. You knocked out everything from the earlier review, including the optional low-priority bits.

I did notice one small follow-up on re-review. I don't think it's worth blocking on, and I'm happy for it to land as a separate PR or issue if that's easier. Just wanted to note it while I had it in my head.

---

**Follow-up: I think `useVocabulary` / `useContextPrefixes` can get stranded in `loading=true` if the URL becomes `undefined` mid-fetch.**

Unless I’m reading this wrong, I think the flow is:

1. User is on a record with an LDE, so `useVocabulary(vocabUrl)` sets `loading=true` and starts a fetch.
2. User navigates in the SPA to a record with no LDE, so `vocabularyUrl` becomes `undefined`.
3. Effect cleanup sets `cancelled = true`. The new effect hits `if (!vocabularyUrl)` at `useVocabulary.ts:14`, clears `labels`, and returns.
4. The in-flight `getCachedVocabulary` promise eventually settles, but all of `.then/.catch/.finally` are guarded on `!cancelled`, so `setLoading(false)` never runs.
5. `RecordDetailPage.tsx:57` ORs the three loading flags together, so the new record can get stuck on `"Loading…"` indefinitely.

Looks like the same pattern exists in `useContextPrefixes.ts:14-18`.

There is also a softer stale-data case: if the URL changes from A → B rather than becoming `undefined`, `labels` keeps A's resolved values until B's fetch comes back, so consumers can briefly see A's labels while B is loading.

I think the fix is probably just to reset a little more aggressively on both the falsy-URL path and the new-fetch path. Something along these lines:

```diff
diff --git a/src/hooks/useVocabulary.ts b/src/hooks/useVocabulary.ts
@@
   useEffect(() => {
     if (!vocabularyUrl) {
       setLabels(null);
+      setError(null);
+      setLoading(false);
       return;
     }
     let cancelled = false;

+    setLabels(null);
     setLoading(true);
     setError(null);

diff --git a/src/hooks/useContextPrefixes.ts b/src/hooks/useContextPrefixes.ts
@@
   useEffect(() => {
     if (!contextUrl) {
       setContext(null);
+      setError(null);
+      setLoading(false);
       return;
     }
     let cancelled = false;

+    setContext(null);
     setLoading(true);
     setError(null);
```

My read is:

* the added resets in the falsy-URL branch make sure a stranded in-flight fetch can't leave the hook pinned in loading 
* the `setLabels(null)` / `setContext(null)` before a new fetch clears stale data during transitions, so consumers don't momentarily see URL A's results while URL B is loading

A tight little test for each hook would probably pin it down nicely too: start a never-resolving fetch (`new Promise(() => {})`), rerender with `undefined`, then assert `loading === false` and data cleared. Feels like the kind of thing that would fail cleanly if someone removed the reset later.

Approving now either way. Up to you whether you want to fold this into the PR or throw it into a follow-up.

PR #19, opened 2026-04-23
Fix storage account name length and SPA 404 on Front Door

danielbreves:
## Summary

- Storage account name `feevidencerepouiproduction` (26 chars) exceeds Azure's 24-character limit — truncates `name_short` to 22 chars so the `fe` prefix stays within bounds
- SPA routes (e.g. `/esea`) return HTTP 404 from blob storage even though the page renders — adds a Front Door URL rewrite rule to serve `/index.html` for extensionless paths so the origin returns 200

### DIFF
--- a/infra/app/locals.tf
+++ b/infra/app/locals.tf
@@ -1,7 +1,7 @@
 locals {
   name          = "${var.app_name}-${var.environment}"
   environment_short = var.environment == "development" ? "dev" : var.environment
-  name_short    = "${replace(var.app_name, "-", "")}${local.environment_short}"
+  name_short    = substr("${replace(var.app_name, "-", "")}${local.environment_short}", 0, 22)
   is_production     = var.environment == "production"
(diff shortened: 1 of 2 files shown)


infra/app/locals.tf
```diff
@@ -1,7 +1,7 @@
 locals {
   name          = "${var.app_name}-${var.environment}"
   environment_short = var.environment == "development" ? "dev" : var.environment
-  name_short    = "${replace(var.app_name, "-", "")}${local.environment_short}"
+  name_short    = substr("${replace(var.app_name, "-", "")}${local.environment_short}", 0, 22)
```
Adam-Hammo: I'm a bit cagey about truncating what essentially is the environment. Could it make sense to have an `app_name_short` (like `feevrepoui`)? 

Or, if it's short enough, we use the four-letter prefixes pretty widely elsewhere (`deve`,`stag`,`prod`) instead of the conditional `environment_short`)

danielbreves: Fair point!

Adam-Hammo (approved):

PR #20, opened 2026-04-24
Add keycloak auth layer

Adam-Hammo:
Adds a simple keycloak layer and repository API plumbing to the UI. Users must login when accessing the UI, and can see their status/sign out in the navbar.

Currently deployed in dev, see for eg: https://data.dev.evidence-repository.org/esea/references/73992e83-629c-49b3-ac83-34f1c7e2a038

Also fixes an issue with custom domain names using 4-letter envs (eg `data.deve...`). Pending applies in terraform should be dismissed, this will bridge the gap.


src/components/layout/AppShell.tsx
```diff
@@ -1,17 +1,26 @@
...
   children: ComponentChildren;
 }
 
 export function AppShell({ children }: AppShellProps) {
+  const { username, logout } = useAuth();
+
   return (
     <div class="app-shell">
       <header class="app-header">
         <a href="/" class="app-header__logo">
```
jaybea: Clicking the `Evidence Repository` header logo after sign-in lands on the `Page not found` screen. The brand link points to `/` but the router only matches `/:community/*`. I think `#22` already addresses this on `issue-7-search-page` (sets `BRAND_HREF = "/esea"` and adds a `useCurrentCommunity()` hook for multi-community future-proofing). 

Both PRs modify `AppShell.tsx` in different ways, so whichever merges second will need a careful resolution that keeps both the community-aware brand link from `#22` and the username display + Sign out button from this PR.

README.md
```diff
@@ -29,6 +29,15 @@ Stack: Vite + Preact + TypeScript
...
 | `VITE_API_BASE` | `/api` | destiny-repository API base URL |
-| `VITE_API_TARGET` | `http://localhost:8080` | Dev server proxy target |
+| `VITE_API_TARGET` | `http://localhost:8000` | Dev server proxy target for `/api` |
+| `VITE_KEYCLOAK_URL` | _(required)_ | Keycloak base URL, e.g. `https://auth.evidence-repository.org` |
+| `VITE_KEYCLOAK_REALM` | _(required)_ | Keycloak realm, e.g. `destiny` |
+| `VITE_KEYCLOAK_CLIENT_ID` | _(required)_ | Keycloak public client ID, e.g. `evidence-repo-ui-client-development` |
+
+## Authentication
+
+The app requires Keycloak authentication on all routes (`onLoad: 'login-required'`). On load, it redirects unauthenticated users to Keycloak and attaches a Bearer token to every destiny-repository API request. You need a Keycloak account in the `destiny` realm to run the app locally.
```
jaybea, line 41: The README implies that having a `destiny` realm account is the gate to running the app locally, but I think there's something else blocking it. I went through the flow: got the account sorted, reset my password, got past the keycloak login redirect, and then the token-exchange POST to `auth.evidence-repository.org` returned 503 from `localhost:3000`. Looks like the Front Door domain allowlist is rejecting non-whitelisted origins (matches the Mar 13 terraform change "only allow keycloak operations from custom domain"). I think the README either needs to say "test against deployed dev for now", or the dev client + Front Door need to explicitly support `localhost:3000`. Either's fine, just want the docs to match what actually works.

src/main.tsx
```diff
@@ -1,6 +1,42 @@
...
+
+function Loading() {
+  return (
+    <div class="auth-gate">
+      <p>Signing you in…</p>
+    </div>
+  );
+}
+
+function AuthError() {
```
jaybea: Optional: Maybe worth adding a test for this branch - mock `initKeycloak` to reject and assert `Unable to sign in.` appears with the retry button. The happy path is well-covered; this is the only un-tested rendering path I noticed.

src/auth/AuthContext.tsx
```diff
@@ -0,0 +1,50 @@
...
+    | { name?: string; preferred_username?: string; email?: string }
+    | undefined;
+  return token?.name ?? token?.preferred_username ?? token?.email;
+}
+
+export function AuthProvider({ children }: { children: ComponentChildren }) {
+  const [, setVersion] = useState(0);
+
+  useEffect(() => {
+    const bump = () => setVersion((v) => v + 1);
```
jaybea, line 25: Optional: Would be helpful to me to have a comment explaining intent here...("force re-read of `keycloak.tokenParsed` on auth events")

jaybea (approved):
No required code changes, just some awareness (PR #22) and a README update?

**Verified live (deployed dev, signed in as me):**

- Login redirect, token exchange, AppShell render
- Bearer-tokened API calls accepted by destiny-repository
- Silent re-auth via SSO cookie on subsequent visits
- Sign-out clears SSO session
- Deep-link preservation through the logout/re-login cycle (`redirect_uri` defaults to current page, so we get this for free!)

**Much Excellence in your Code**

- `AuthProvider` outermost in `App.tsx` so `AppShell` can use `useAuth()`.
- `Loading` / `AuthError` boundary in `main.tsx`. Explicit retry button rather than silent breakage.
- `keycloak.init` uses PKCE `S256` and `checkLoginIframe: false`. This is the way to go for modern SPAs (the iframe-based session check is broken under third-party cookie restrictions anyway).
- Centralised mock in `tests/setup.ts` with per-test reset. No per-file mocking ceremony, yay!
- `domain_env_label` map in `infra/app/locals.tf`. Cleaner than `substr(var.environment, 0, 4)` and explicit about the truncation rules.

1. **Coordination with #22 on `AppShell.tsx`** — both PRs modify the same file in different ways (this PR adds username + sign-out; #22 adds a community-aware brand link with a `useCurrentCommunity()` hook). Whichever lands second will conflict; the resolution needs to keep both. The `<a href="/">` brand link in this PR also lands on `NotFound` after login because the router only matches `/:community/*`, but #22 already addresses that with `BRAND_HREF = "/esea"`. No code change needed here if the merge sequence is planned, just flagging.

2. **README vs reality on local auth.** The README reads as if local Keycloak auth should work, but the token endpoint POST returns 503 from `localhost:3000`. Looks like the Front Door domain allowlist (the Mar 13 terraform change "only allow keycloak operations from custom domain"). Either the docs should say "test against deployed dev for now", or the dev client + Front Door need explicit `localhost:3000` support. If local does end up being supported, `vite.config.ts` should also set `strictPort: true`. Without it, a port collision causes Vite to silently shift to `3001`, but Keycloak's redirect URI whitelist is pinned to `:3000`, so the post-login redirect would land at whatever else is on `:3000`. Failing loudly at startup is safer.

src/components/layout/AppShell.tsx
Adam-Hammo: Nice, good catch - let's make sure it lands correctly

README.md
Adam-Hammo: Good callout - it would work with a (magical and yet unspecified) local keycloak, but not the deployed instance. I've actually updated https://github.com/destiny-evidence/destiny-shared-infra/pull/33 to allow `localhost` as a web origin in dev/staging, so locally we can authenticate against those now and the README is accurate. Long term we might want to find a nice solution to spinning up the "entire" environment (keycloak, repo, evrepo-ui, other apps) as one.

Issue #21, opened 2026-04-24
Link concept tags to taxonomy definitions


danielbreves:
## Context

[Issue #9](https://github.com/destiny-evidence/evidence-repo-ui/issues/9#issuecomment-4205588338) calls for concept tags on finding cards to be "linkable to taxonomy definitions":

> Concepts should display with their property label (e.g. "education level: Lower Primary") and be linkable to taxonomy definitions.

## Blocker

The taxonomy-builder has an open issue for a URI resolution service: [destiny-evidence/taxonomy-builder#63](https://github.com/destiny-evidence/taxonomy-builder/issues/63). It plans a `GET /api/resolve?uri={uri}` endpoint returning concept label, definition, scheme, and broader/narrower.

## Proposed follow-up

Once #63 ships:
1. Decide the target URL — either the resolve endpoint directly or a taxonomy-builder UI route that uses it under the hood.
2. Extend `HierarchicalTag` / `TagGroup` in `src/components/TagGroup.tsx` to accept a `uri` and render as a link when present.
3. Build the link in `conceptsToTags` (`src/services/conceptLabels.ts`) from each concept's URI.
4. Style the linked tag state (hover, focus) without changing the default ghost-tag visuals.

Issue #10, opened 2026-04-08
Record detail page: Effect estimate and arm data cards


danielbreves:
Render effect estimates and arm data within each finding card. Layout left to implementer — the EV Core class hierarchy in the PRD provides design direction.

### Available data per effect estimate (from SHACL shapes)

- pointEstimate, standardError, variance, confidenceLevel, CI lower/upper
- effectSizeMetric (concept label, e.g. "Hedges' g", "Cohen's d")
- estimateSource (concept label, e.g. "Computed from summary statistics")
- baselineAdjusted, clusteringAdjusted
- derivedFrom → ObservedResult, forOutcome → Outcome

### Available data per arm (ObservedResult)

- forCondition → Condition (intervention or control, via blank node ref)
- n, mean, sd, se, preMean, preSd, clusterCount, icc, events

mootpointer:
## Design Spec

**Prototype:** [Search page](https://taxonomy-prototype.pages.dev/demo/search) · [Investigation detail](https://taxonomy-prototype.pages.dev/demo/investigation)

### Effect estimate card

Nested inside each finding card, with `margin-top: 16px` to separate from the outcome/measure section above. Inset background (`#F8F9FA`), 1px border, 4px radius.

**Layout, in order:**

1. **Label (h4):** "EFFECT ESTIMATE" — Söhne Mono 10px, uppercase, weight 400, tertiary color
2. **Headline row:** Point estimate (Söhne Mono 30px, 700 weight) + metric label (Söhne Mono 13px) + SE (Söhne Mono 13px tertiary). All on one baseline-aligned flex row.
3. **CI text:** "CI [lower, upper]" — Söhne Mono 14px, brackets in tertiary color. Don't assume "95% CI" — the data doesn't specify confidence level.
4. **CI visualisation** (see below)
5. **Adjustment tags:** Ghost outlined style (`transparent` bg, 1px `#D8DCE2` border, 3px radius) — "Baseline adjusted", etc.
6. **Negative indicators as text:** "Not clustering adjusted" as plain 12px tertiary text when `clusteringAdjusted` is `"no"`

### Effect size coloring

The point estimate number uses restrained coloring:
- **Positive, CI does not cross zero:** Primary text color (`#161B22`)
- **Negative, CI does not cross zero:** Muted dark red (`#6B2020`)
- **CI crosses zero (non-significant):** Secondary text color (`#424A53`)

No bright green/red — the CI visualisation communicates significance.

[...]

Issue #9, opened 2026-04-08
Record detail page: Finding cards


danielbreves:
Render one card per finding within the record detail page. Card layout and grouping of fields is left to the implementer — the EV Core class hierarchy in the PRD provides design direction, but the final presentation should emerge from working with the data.

### Available data per finding (from SHACL shapes)

- Intervention: name, education themes, implementer type, implementation description, fidelity, duration, funder
- Context: education levels, settings, country, country level 1, participants
- Outcome: outcome concept labels with supporting text
- Sample features, sample size, attrition, cost, group differences
- Supporting text on coding annotations

Concepts should display with their property label (e.g. "education level: Lower Primary") and be linkable to taxonomy definitions.

Multi-arm studies are particularly important to render well — clearly show which arms are being compared.

mootpointer:
## Design Spec

**Prototype:** [Search page](https://taxonomy-prototype.pages.dev/demo/search) · [Investigation detail](https://taxonomy-prototype.pages.dev/demo/investigation)

### Shared context block

When multiple findings reference the same intervention, control, context, and sample (via blank node IDs), these are factored out into a single shared block above the finding cards.

**Structure:**
- White card, 4px radius, subtle shadow
- Header (h2): "SHARED ACROSS ALL FINDINGS" — Söhne Mono 12px, uppercase, primary color
- **Comparison row:** 3-column grid — Intervention | "vs" | Control
  - Each side has an h3 label ("INTERVENTION" / "CONTROL") and a Signifier 16px value
  - Intervention includes labelled tag group for education theme, with hierarchy inside the tag
  - Intervention description as a left-bordered blockquote (2px border, 14px left padding)
  - Duration as Söhne Mono text

### Context and Sample sections

Displayed as separate h3-level sections ("CONTEXT", "SAMPLE"), each using a **two-column grid** for field groups:

```
CONTEXT
LEVEL    Primary Education     SETTING      Formal
PARTICIPANTS  Students

SAMPLE
SIZE  667                      FEATURES     SES
```

This uses `display: grid; grid-template-columns: 1fr 1fr; gap: 6px 24px;` to flow attributes across both columns and avoid wasting horizontal space.

### Hierarchy inside tags

For concepts with a parent in the SKOS hierarchy (via `skos:broader`), show the parent as a faded prefix inside the tag:

```
[School Organization › Tracking of Students]
```

[...]

PR #22, opened 2026-04-26
Issue #7: Search page and results list

### LINKED ISSUE #7: Search page and results list

jaybea:
## Summary
- Realign the search page, results list, and reference detail page to Andrew's klim-b prototype variant per #7
- Five tier passes landed as separate commits: 920px editorial column → hero + search bar → result row 1fr/auto grid → pagination + header brand → InvestigationCard venue/DOI polish
- Klim font binaries integrated under the project's Web licence; result-count emphasis recovered via mono typography rather than the prototype's bold weight
- Closes #7

## Validations performed
- 195 unit tests passing across 21 files (`vitest run`)
- Visual comparison against the klim-b prototype (`/demo/search-klim-b` and `/demo/investigation-klim-b`) for hero, results list, pagination, InvestigationCard
- InvestigationCard verified on live data across LDE variants: standard Journal Article, Thesis/Dissertation, retracted state (pink banner + 50% opacity), mojibake-titled record (renders without crash), missing-DOI record (DOI line cleanly omitted)
- Vocabulary-unavailable fallback verified — italic tertiary "Vocabulary unavailable — some labels could not be resolved." replaces the doc-type pill; divider preserved
- Venue + pagination rendering verified by injecting synthetic data into a bibliographic enhancement (restored after capture); confirmed italic Signifier `Journal of Applied Developmental Psychology, 31(4), 304–320` idiom
- Pagination row matches prototype at 34×34 outlined buttons, filled-accent active, mono numerals, ellipsis between near pages and last page
- Year-range filter label re-rendered in plain mixed-case sans tertiary to match prototype (was mono uppercase)
- Community swap reactivity verified with a throwaway second community (popstate hook on `useCurrentCommunity`)
- All seven licensed Klim cuts loaded and rendering on live pages (verified via `document.fonts` API and computed-style probes); production build bundles each woff2 with cache-busting hashes
- Fallback path verified by stripping Klim from the CSS font-family variables at runtime — page renders cleanly via Georgia + system sans + SFMono on both search and detail routes, no layout breakage
- `font-synthesis: none` confirmed in effect — Signifier `.row-title` (declared `font-weight: 500`) renders Signifier Regular instead of a synthesised bold

## Fonts — Klim Type Foundry licensed binaries
The seven licensed cuts (Signifier Regular + Italic, Untitled Sans Medium + Italic, Söhne Mono Buch/Kräftig/Halbfett) live at `src/styles/fonts/` and are wired into `src/styles/fonts.css` as static-cut `@font-face` declarations. Licence terms are summarised at `licenses/klim/README.md` (Klim Web licence — 5,000 MAU on evidence-repository.org; standard terms at klim.co.nz/licences). Proof-of-purchase is held off-repo.

`font-synthesis: none` is set on `<body>` so the browser uses the nearest declared cut from the same family rather than synthesising weights or italics outside the licensed set. Any CSS asking for Signifier at 500 or 700 (e.g. result-row title) renders Signifier Regular at the correct font-family, not a faux-bold.

Fallback chains (for environments where the binaries fail to load):
- display/serif → Georgia → "Times New Roman" → serif
- body → `-apple-system` → `BlinkMacSystemFont` → "Segoe UI" → Roboto → sans-serif
- mono → SFMono-Regular → Menlo → Monaco → monospace

## Notes for design review
The Klim Web licence covers two cuts each of Signifier and Untitled Sans and three of Söhne Mono — fewer weights than the prototype's variable-font CSS expects. With `font-synthesis: none`, requests for unlicensed weights resolve to the closest licensed cut from the same family. Three places where the prototype's typographic intent is only partially carried:

- **Body text reads at Untitled Sans Medium**, not Regular. Untitled Sans Regular isn't licensed; Medium is declared under `font-weight: normal` so default body text matches it directly. I think the page reads better at this scale than the prototype's Regular — denser per line, slightly more grounded — and am calling this an intentional divergence rather than a compromise. Easy to revisit.
- **Hero subtitle italic** renders Signifier Regular Italic, not Light Italic (300 is not licensed). Slightly less "flowery" than the prototype.
- **Result-row title and Search button label** render at the licensed Medium-weight cuts of their families. The prototype declared 500/600 respectively. Hierarchy is preserved by size, colour, and (for the button) the filled accent background.

The prototype's bold result count (`**47** results for "phonics instruction reading"`) is recovered via typography rather than weight: the count is wrapped in a span styled `var(--font-mono) 15px var(--color-text-primary)` — Söhne Mono Buch at slightly larger size and primary colour. Quotes around the query echo are now curly doubles for editorial polish.

If we'd like to close any of those gaps, the additional cuts to license are:
- **Signifier Light Italic** (300 italic) — restores the hero subtitle's lighter italic
- **Signifier Medium** (500 roman) — restores result-row title weight contrast vs. the abstract preview
- **Untitled Sans Bold** (700 roman) — restores Search button label emphasis (and any future pill labels at weight ≥ 600)

## Out of scope (covered by other tickets or deferred)
- Concept pills on result rows — #21 (blocked on taxonomy-builder #63)
- Sort dropdown — deferred, no spec
- Stat-badge real values rendered as `—` placeholders — deferred; needs backend `findings_count` / `estimates_count` in search response
- Investigation page below the header card (shared block, finding cards, effect estimates, arm data) — #9 and #10


src/components/search/SearchBar.tsx
```diff
@@ -0,0 +1,109 @@
...
+interface SearchBarProps {
+  q: string;
+  startYear: number | undefined;
+  endYear: number | undefined;
+  onSubmit: (q: string, startYear: number | undefined, endYear: number | undefined) => void;
+  disabled?: boolean;
+}
+
+function parseYear(raw: string): number | undefined {
+  // Strict decimal only — matches parseSearchParams' rule so URL state and
```
Adam-Hammo: Could/should this be a shared function in a `utils` somewhere so both searchParams and searchBar can use the same mechanism?

src/pages/SearchPage.tsx
```diff
@@ -7,14 +17,169 @@ interface SearchPageProps {
   community?: string;
 }
 
+// Mirrors the destiny-repository /v1/references/search/ fixed page size.
+// Not a UI knob — changing this here desyncs totalPages math from the
+// server's actual paging. If the backend ever exposes a configurable page
+// size, switch to reading it off SearchResult.page rather than tuning here.
+const BACKEND_PAGE_SIZE = 20;
```
Adam-Hammo: Each API response returns `["page"]["count"]` which tells you the size of the page - is it easy to use this, like what `formatTotal` does below?

src/pages/SearchPage.tsx
```diff
@@ -7,14 +17,169 @@ interface SearchPageProps {
...
+  useEffect(() => {
+    const current = search.startsWith("?") ? search.slice(1) : search;
+    if (current !== canonicalQs) {
+      navigate(buildSearchUrl(community.slug, params), { mode: "replace" });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [canonicalQs, community.slug]);
+
+  const corpus = useCorpusTotal(community);
+  const results = useSearch(params, community.defaultAnnotations);
```
Adam-Hammo, line 73: I'd be interested in exploring a community context provider that we can use throughout the UI to help scope things like this. In particular, this feels like an eventual footgun, where community-specific information has to be threaded into each call. Probably not for this PR though. What do you think?

src/pages/SearchPage.tsx
```diff
@@ -7,14 +17,169 @@ interface SearchPageProps {
...
 
   return (
     <div class="search-page">
-      <h1>Search {community.name}</h1>
-      <p>Search page for {community.name}.</p>
+      <section class="search-hero">
+        <h1 class="search-hero__title">Search the evidence</h1>
+        <p class="search-hero__subtitle">
+          {corpus.total
+            ? `${formatTotal(corpus.total)} investigations across ${community.name.toLowerCase()} research`
```
Adam-Hammo, line 110: I fear we'll exceed 10,000 investigations - this will be good enough for now but we might need to consider how enhancing the API to get a summative number somehow. Perhaps our first ES aggregation via API?

Adam-Hammo (approved):
Two nits and two thought-provokers, nothing showstopping. Nice stuff!

src/pages/SearchPage.tsx
jaybea: filed as #27 for follow-up after the merge dust settles

Adam-Hammo (approved):

PR #23, opened 2026-04-27
Add finding cards with shared context block

### LINKED ISSUE #9: Record detail page: Finding cards

danielbreves:
Closes #9.

## Summary

Renders one card per finding on the record detail page with intervention, control, context, outcome, and sample details. When multiple findings share the same intervention/control/context (via blank-node IDs), a "Shared across all findings" block is factored out above the finding cards. Hierarchical concept tags (parent › child via `skos:broader`) are resolved from the vocabulary's JSON-LD, and tags with a `skos:definition` show that definition as a hover tooltip.

Effect estimates and arm data are intentionally deferred to #10.

## Highlights

- Two-pass JSON-LD parser handles blank-node references (`evaluates: "_:intervention"` resolving to a full object defined in another finding)
- Vocabulary service extended to extract `skos:broader` for hierarchical tags and `skos:definition` for tooltip text
- Per-section "Source evidence" toggles with structured panels (label + `p. N:` page prefix extracted from supportingText)
- Concept tags with definitions render a CSS-only dark tooltip on hover/focus-visible (`aria-controls` / `aria-expanded`, keyboard-accessible via `tabIndex={0}`)
- Retracted records dim the findings section alongside the investigation card
- Graceful degradation when vocab/context fetch fails: findings still render with raw concept identifiers and a banner explains the degraded state (mirrors PR #18's investigation-side handling)
- Shared CSS utilities (`.lg-card`, `.lg-divider`, `.lg-field-grid`, `.lg-field`, `.lg-label`, `.lg-section-label`, `.lg-kicker`, `.lg-retracted`) live in `variables.css` so cards/labels stay consistent
- Reusable layout components: `ComparisonRow`, `LabeledField`, `SourceEvidenceToggle`, plus per-section detail components (`InterventionDetails`, `ContextDetails`, `SampleDetails`, `OutcomeDetails`)

## Multi-arm framing

Issue #9 calls out multi-arm studies as needing clear rendering. When findings differ on intervention/control/context, `findSharedContext` returns `null` and every card falls back to inline rendering of its own intervention/context — crude but correct as an MVP. A proper multi-shared-block layout (stack-N, group-by-arm, etc.) is a follow-up design question and should be tracked as a separate issue.

## Deferred / follow-ups

- Effect estimates + arm data (#10)
- Linkable concept tags (#21, blocked on taxonomy-builder URI resolver)
- Multi-shared-block layout for multi-arm studies (separate follow-up)

## Test plan

- [x] `npm test` — 119 tests pass
- [x] Load `/:community/references/0196b1a0-0000-7000-8000-000000000010` — multi-finding record renders shared context block + 3 finding cards, each card shows the back-reference and its own outcome/sample
- [x] Load a single-finding record — no shared block, finding card with inline intervention/context/sample
- [x] Load a record without linked data — no regression on investigation card
- [x] Load the retracted record — investigation card and findings section both dim to 0.5 opacity
- [x] Open and close Source evidence toggles — `aria-expanded` flips, panel hidden state toggles, `aria-controls` always points at a real DOM element
- [x] Hover (or tab to + focus) a concept tag whose vocabulary entry has `skos:definition` — dark tooltip appears above the tag
- [x] Simulate vocab fetch failure (e.g. break `vocabulary_uri`) — findings still render with raw concept identifiers and a "Vocabulary unavailable" banner appears above them


src/components/SourceEvidenceToggle.tsx
```diff
@@ -0,0 +1,63 @@
...
+  if (entries.length === 0) return null;
+
+  return (
+    <div class="source-evidence">
+      <button
+        class="source-evidence__toggle"
+        type="button"
+        onClick={() => setOpen(!open)}
+        aria-expanded={open}
+        aria-controls={panelId}
```
jaybea, line 39: `aria-controls={panelId}` is set unconditionally, but `<div id={panelId}>` only mounts when `open` is true. ARIA contract requires the referenced element to exist in the DOM. Standard fix is to render the panel always and toggle visibility with `hidden`:

```tsx
<button
  class="source-evidence__toggle"
  type="button"
  onClick={() => setOpen(!open)}
  aria-expanded={open}
  aria-controls={panelId}
>
  <span class="source-evidence__arrow">{open ? "▾" : "▸"}</span>{" "}
  Source evidence
</button>
<div id={panelId} class="source-evidence__panel" hidden={!open}>
  {entries.map((entry, i) => {
    ...
  })}
</div>
```

src/components/ComparisonRow.css
```diff
@@ -0,0 +1,28 @@
...
+.comparison-row__condition {
+  min-width: 0;
+}
+
+.comparison-row__condition-label {
+  margin: 0 0 var(--spacing-xs);
+}
+
+.comparison-row__condition-value {
+  font-size: 16px;
```
jaybea: `16px` here is the same value as `var(--font-size-base)` (defined in `variables.css:11` as `1rem`). Two lines below at line 24, the file uses `var(--font-size-sm)` correctly - looks like a slip rather than a deliberate exception.

```suggestion
  font-size: var(--font-size-base);
```

src/components/ContextDetails.tsx
```diff
@@ -0,0 +1,68 @@
...
+    labels,
+    broader,
+    definitions,
+  );
+  const participantTags = context.participants?.map((p) => p.value) ?? [];
+  const evidenceEntries = collectEvidence(context);
+
+  return (
+    <>
+      <div class="context-details__grid lg-field-grid">
```
jaybea: `context-details__grid` / `__field` / `__field-label` are declared on these elements but no `ContextDetails.css` file exists and nothing in the codebase selects `context-details__*`. The `.lg-field-grid` / `.lg-field` / `.lg-label` utilities here do all the styling work, so visually nothing breaks - but the BEM classes are dead.

Compare with `InterventionDetails`, `SampleDetails`, `OutcomeDetails`, each of which imports its own CSS module. Either add a `ContextDetails.css` to match the sibling pattern or drop the dead BEM classes.

src/services/vocabulary/vocabularyService.ts
```diff
@@ -20,48 +28,66 @@ function toJsonLdUrl(vocabularyUrl: string): string {
   return url.toString();
 }
 
+function extractBroaderUri(
+  value: JsonLdGraphEntry["skos:broader"],
+): string | undefined {
+  if (!value) return undefined;
```
jaybea, line 34: SKOS allows polyhierarchy - a concept can have multiple `skos:broader` parents. This silently picks the first and drops the rest. Probably the right UX call for breadcrumb display (multiple parent chains in a single tag would be visually messy), but worth a 1-line comment for the next reader:

```suggestion
  // SKOS allows polyhierarchy; we surface only the first broader for breadcrumb display.
  const first = Array.isArray(value) ? value[0] : value;
```

src/components/FindingsSection.tsx
```diff
@@ -0,0 +1,51 @@
...
+          shared={shared}
+          sampleSource={findings[0]}
+          labels={labels}
+          broader={broader}
+          definitions={definitions}
+        />
+      )}
+      {findings.map((finding, i) => (
+        <FindingCard
+          key={i}
```
jaybea: Same `key={i}` observation that came up on PR #18's `TagGroup`. Lists here are immutable so the bug class doesn't trigger, but stable keys are cheap. Recurs in `SourceEvidenceToggle.tsx:49` too. Flagging for consistency with the prior review - no action required if you'd rather not chase it.

jaybea (changes requested):
Solid parser foundation. The two-pass blank-node resolution is the right shape for the LDE pattern, and the `skos:broader` / `skos:definition` extensions fit in cleanly. The bibliographic header and retracted handling also preserve PR #18's invariants.

Two blockers below, plus an accessibility issue inline.

### Bug 1 (must fix) - `SharedContextBlock` hoists sample data on weaker proof than it actually has

`groupFindings()` ([`findingGroups.ts:33-79`](https://github.com/destiny-evidence/evidence-repo-ui/blob/d751d04/src/services/findingGroups.ts#L33-L79)) only proves that intervention / control / context IDs match across all findings. Its return type `SharedContext` reflects that, with three fields and no sample. But [`FindingsSection.tsx:28`](https://github.com/destiny-evidence/evidence-repo-ui/blob/d751d04/src/components/FindingsSection.tsx#L28) then goes a step further and passes `sampleSource={group.findings[0]}` into `SharedContextBlock`, and [`FindingCard.tsx:80-90`](https://github.com/destiny-evidence/evidence-repo-ui/blob/d751d04/src/components/FindingCard.tsx#L80-L90) suppresses `<SampleDetails>` whenever `isShared` is true. I think the use site is assuming more than the prover has actually established.

Two consequences, both real, I think:

- `finding[0]` gets the full intervention / control / context / sample rendered inline because `isShared = group.shared !== null && i > 0` ([`FindingsSection.tsx:39`](https://github.com/destiny-evidence/evidence-repo-ui/blob/d751d04/src/components/FindingsSection.tsx#L39)) excludes index 0. So every shared-context record renders the shared block once, then immediately repeats it inside Finding 1's card. I reproed this live on `/esea/references/019d986a-9138-761b-a187-516f6524a029` (Savage 2009 ABRACADABRA RCT, 3 findings).
- `findings[k>0]` have their sample sections suppressed even though nothing has actually proved that sample is shared. Right now that is a bit camouflaged because the parser does not yet resolve `sampleSize` blank-node refs, so later findings happen to parse sample as `null` anyway. But I think this turns into a visible correctness bug the moment sample-side ref resolution lands, or as soon as a record carries per-finding sample inline.

This cuts across Issue #9's contract pretty directly. Sample features, sample size, attrition, cost, and group differences are all called out there as per-finding data.

Fix direction looks pretty straightforward to me:

- Drop `sampleSource` from `SharedContextBlock`'s API, and render `<SampleDetails>` in every `FindingCard`.
- Change `isShared` to `group.shared !== null` so `finding[0]` also shows the back-reference lozenge instead of duplicating the shared block.

I think the test that let this through is just a bit too presence-based. [`FindingsSection.test.tsx`](https://github.com/destiny-evidence/evidence-repo-ui/blob/d751d04/tests/components/FindingsSection.test.tsx) asserts that "Intervention appears more than once", which passes whether it appears 2 times or 4+ times. Worth tightening that to something exclusive, for example: "`InterventionDetails` does not render inside `FindingCard` at index 0 when shared".

### Bug 2 (must fix) - vocab/context fetch failure silently removes the entire findings section

PR #18 already hit this failure mode for `InvestigationCard` and fixed it by decoupling `extractIsRetracted` from vocab and surfacing a `vocabUnavailable` flag. I do not think the same lesson carried into the findings path.

In [`RecordDetailPage.tsx:40-43`](https://github.com/destiny-evidence/evidence-repo-ui/blob/d751d04/src/pages/RecordDetailPage.tsx#L40-L43), if either `useVocabulary` or `useContextPrefixes` errors, `investigation` stays `null` and the entire findings region disappears. No warning, no signal, just gone. So the user gets the `InvestigationCard`'s "Vocabulary unavailable" footnote and then a blank space where the findings should be.

I think this wants the same graceful-degradation pattern as #18:

- Pass `labels: new Map()` into `parseInvestigation` as a fallback, and same idea for `broader` / `definitions`.
- Render the findings with raw URIs as fallback labels when vocab resolution is unavailable.
- Surface a `vocabUnavailable` banner above the section.

That at least preserves the findings content instead of making the whole section vanish.

### Multi-arm MVP framing

When findings do not all match on intervention / control / context, `groupFindings` returns `shared: null` and falls back to inline rendering in every card. For multi-arm records with multiple distinct intervention groups, I think this is crude but safe.

Issue #9 does call out multi-arm studies as something that needs to be rendered clearly, so I would not treat this as the finished answer. But as an MVP fallback, I think it is defensible. Probably worth calling that out in the PR description so reviewers do not read it as an oversight.

Also worth a follow-up issue for the actual multi-shared-block layout question, whether that ends up being stack-N, group-by-arm-group, or something else.

### Other

Bug 3 (accessibility, must fix) and Bug 4 (CSS token, modest) inline.

Three nits at code-comment level as well. I flagged them inline, but none of those are merge-blocking.

jaybea:
I realised the Klim-licensed fonts are coming in PR #22 , and these two PRs are going to have to come together at some point, so I sanity-checking the typography here in #23.

I compared a live LDE-bearing record (de Graaff et al. 2009, *Benefits of systematic phonics instruction*) using PR #23 against the prototype by diffing computed DOM styles back to the prototype CSS. I think the drift looks pretty limited. I think it comes down to a few small clusters, and they only really show up once #22 (Klim integration) is in the picture.

- **Mono label weight:** `.lg-label`, `.lg-section-label`, `.lg-kicker`, and `.investigation-card__kicker` are set to `font-weight: 500` (Söhne Mono Kräftig), whereas the prototype uses `600` (Halbfett). Both cuts are licensed, so this is not a licensing problem. To my eye it just reads a bit lighter than the prototype.
- **Missing serif family:** `.comparison-row__condition-value` and `.investigation-card__venue` do not declare a `font-family`, so they currently just inherit the body text styling. In the prototype both use `var(--font-display)` (Signifier): `font-weight: 500` plus italic for the condition value, and `font-weight: 300` for the venue. I think this one probably does want fixing, since it changes the type treatment rather than just nudging weight.
- **Body weight 500 on Untitled Sans:** `.source-evidence__toggle` and `.sample-details__field-value` ask for `font-weight: 500`. We only purchased Untitled Sans Medium, which only ships at 400, so with `font-synthesis: none` from #22 these quietly fall back. I think the simplest fix is probably just to drop the `500` declaration there. The existing visual cues already seem to carry enough of the distinction anyway: chevron + colour for the toggle, and label/value family contrast for the sample details.

I don’t see any licence violations here. Everything looks to be either drift between licensed cuts or a missing-family rule, rather than anything trying to use an unlicensed weight.

danielbreves:
Thanks for the comprehensive review @jaybea! I think I've addressed all of it.

danielbreves:
> font-weight: 500 plus italic for the condition value

The condition value doesn't look italic to me. Could you tell me where are you seeing this?

Screenshot 2026-04-28 at 1 27 24 PM

jaybea:
My mistake on the italic, sorry about that!

I went back and checked and italic wasn't in the prototype, there's just:
```css
.comparison-group .comp-value {
  font-family: var(--font-display); font-size: 16px; font-weight: 500;
  line-height: 1.3; margin-bottom: 8px;
}
```

I'm claiming font fatigue! But where your change landed, with  (font-family: var(--font-display), font-weight: 500, no italic) is the right end result.

While I was double-checking I caught a second slip in the same bullet that's worth flagging: I told you the prototype venue uses `var(--font-display)` at `font-weight: 300`. It doesn't. The prototype rule for `.investigation-card .inv-venue` is just `font-size: 13px; font-style: italic; color: var(--text-tertiary); margin-bottom: 4px;`. No font-family, no font-weight. So strictly, it inherits Untitled Sans body at default weight with italic. 

That said, what you implemented for the venue (Signifier + weight 300) is probably the right call anyway. Our licensed Untitled Sans only ships the roman cut, and once `font-synthesis: none` is on, "italic Untitled Sans" can't render without faux-italic. Switching to Signifier (which has a licensed italic) is the only way to keep that italic treatment visible. Happy to leave it as you have it, just wanted to be straight with you that the prototype-fidelity framing in my review was off.

Edit:
Venue typography looks great in place. (Synthesized the journal name locally, most DR records don't carry publication_venue yet, so this was the only way to see it rendered.)

image

jaybea (approved):
Thanks for the thorough response @danielbreves. All five inline items addressed cleanly, plus the typography alignment. I think it looks gorgeous!

I REALLY look forward to seeing this in staging, sorry I didn't catch your respones and approve in time for this mornings meeting so the group could see it. :(

And  nice catch on the per-finding `SampleDetails` fix in "Render SampleDetails per-finding; show back-ref on every shared card"!  I missed that. Verified visually on Slavin 1984's TAI study: three findings render distinct sample sizes (504, 451, 1363), where before you fix would have been just "504" three times. The `hasSampleData()` guard reads cleanly too.

`findSharedContext` returning `SharedContext | null` is also a much cleaner API than the old `groupFindings`, and the vocab/context graceful-degradation pattern in "Render findings with identifier fallback when vocab/context fetch fails" mirrors PR #18 nicely.

Heads-up for the rebase: trial-merging `origin/main` shows one trivial conflict in `InvestigationCard.css` line 45 (the `--font-display` fallback) plus three benign auto-merges (`InvestigationCard.tsx`, `variables.css`, `InvestigationCard.test.tsx`). 

PR #22's plain `var(--font-display)` resolves cleaner since the `--font-serif` fallback is now redundant once the Klim tokens land . Both lines produce the same rendered CSS, so picking either is fine.

PR #31, opened 2026-04-29
Issue #27: extract CommunityProvider for shared community lookup

### LINKED ISSUE #27: Refactor: introduce useCommunity() context provider to remove findCommunity duplication
Following up on [Adam's review comment on #22](https://github.com/destiny-evidence/evidence-repo-ui/pull/22#discussion_r3151277480), where he suggested a community context provider to avoid threading community-specific information into every call:

> I'd be interested in exploring a community context provider that we can use throughout the UI to help scope things like this. In particular, this feels like an eventual footgun, where community-specific information has to be threaded into each call.

## Why

Three sites currently look up the community independently:

- `src/components/layout/AppShell.tsx` (header brand)
- `src/pages/SearchPage.tsx`
- `src/pages/RecordDetailPage.tsx`

Each calls `findCommunity(slug)`, handles the undefined case, and threads parts of the community object into hooks (`useCorpusTotal`, `useSearch`) and components (`ResultRow`). As more pages get added, every new one has to remember the lookup pattern.

## Proposed shape

- New `src/community/CommunityContext.tsx` with `CommunityProvider` and a `useCommunity()` hook.
- Provider listens to `URL_CHANGE_EVENT` and `popstate`, derives the community from the first path segment, exposes via context.
- Wrap `<Router>` in `<CommunityProvider>` in `App.tsx`.
- AppShell, SearchPage, RecordDetailPage drop their `findCommunity` calls and use `useCommunity()`.
- Pages still handle `null` by rendering `NotFoundPage` (no rendering changes).

## Design decision

Hooks like `useCorpusTotal(community)` and `useSearch(params, defaultAnnotations)` keep their explicit args rather than consume the context directly. Keeps hooks pure and easier to test in isolation. The win at the page level is the real goal.

## Scope

- 1 new file (provider + hook + tests)
- 4 files modified (`App.tsx`, `AppShell.tsx`, `SearchPage.tsx`, `RecordDetailPage.tsx`)
- Net negative line count

## Out of scope

- Multi-community switcher UI (separate concern, needs design).
- Functional change. Same routing, same NotFound behavior.

## Timing

Should wait for #20, #22, and #23 to merge before basing. Touches `AppShell.tsx` (modified by both #20 and #22) and `SearchPage.tsx` (modified by #22). Picking it up before they all settle is asking for merge pain.

## Testing

A throwaway-second-community pattern works well: temporarily register a second `Community` pointing at the same backend annotation, navigate to the new slug, confirm routing and component rendering behave as expected, then revert before commit. Useful for exercising community-agnostic paths that single-community testing can't reach.

Plus the standard cases: invalid slug renders `NotFoundPage`, deep-link reload initializes the provider from URL state correctly.

Unit tests on the provider should mock `findCommunity` and `window.location.pathname`.

jaybea:
Closes #27.

## Summary
- New `src/community/CommunityContext.tsx` with `CommunityProvider` and `useCommunity()`. Provider derives the community from the URL's first path segment and subscribes to `popstate`, `URL_CHANGE_EVENT`, and `<Router onChange>`.
- AppShell, SearchPage, and RecordDetailPage drop their per-call `findCommunity()` lookups for the shared context. NotFound rendering on null community is unchanged.
- `useCommunity()` throws when used outside `CommunityProvider` so missing-provider bugs surface loudly instead of rendering NotFound silently.

## Notes
- **Anchor-click bridge:** preact-router's global click handler calls `history.pushState` directly without firing `popstate` or our `URL_CHANGE_EVENT`. `<Router onChange>` now dispatches `URL_CHANGE_EVENT` so `<a href="/esea">` clicks stay in sync with the provider. This is load-bearing for `NotFoundPage`'s "Go to search" link and the AppShell brand link, both of which are plain anchors. Same fix incidentally protects `useUrlParams` from the same class of staleness.

## Validations performed
- 271 unit tests passing across 33 files (`vitest run`)
- `npm run typecheck` clean; `npm run build` clean
- New regression test in `tests/App.test.tsx`: starting at `/banana`, clicking "Go to search" navigates to the search page (would fail without the `Router onChange` bridge)
- New test in `tests/community/CommunityContext.test.tsx`: `useCommunity()` outside `CommunityProvider` throws
- Manual walkthrough on localhost: golden path, search submit, pagination, click-through to detail, browser back/forward, cold deep-link, `/` (no community), `/banana` (invalid slug), anchor-click bridge from NotFound, invalid reference id. Brand display followed the community-on/off rules at every step; search annotation pass-through (`domain-inclusion/jacobs-education`) verified in Network tab.

## Out of scope
- Multi-community switcher UI.
- Pre-existing dev-only sign-out redirect issue: signing out from localhost hits "Invalid redirect uri" because Keycloak admin needs `http://localhost:3000/*` added to **Valid Post Logout Redirect URIs** on `evidence-repo-ui-client-development`. Staging client verified correctly configured.

### DIFF
--- a/src/pages/SearchPage.tsx
+++ b/src/pages/SearchPage.tsx
@@ -68,8 +68,8 @@ function SearchPageInner({ community }: { community: Community }) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [canonicalQs, community.slug]);
 
-  const corpus = useCorpusTotal(community);
-  const results = useSearch(params, community.defaultAnnotations);
+  const corpus = useCorpusTotal();
+  const results = useSearch(params);
 
   // Meta-bar appears whenever the user has narrowed from browse mode (a query
   // or a year filter), or whenever an error needs surfacing. Empty q + no
--- a/src/pages/SearchPage.tsx
+++ b/src/pages/SearchPage.tsx
@@ -1,5 +1,5 @@
 import { useEffect } from "preact/hooks";
-import { findCommunity } from "@/services/communities";
+import { useCommunity } from "@/community/CommunityContext";
 import type { Community } from "@/types/models";
 import { parseSearchParams, toQueryString, buildSearchUrl } from "@/services/searchParams";
 import { navigate } from "@/services/navigation";
@@ -14,7 +14,6 @@ import "./SearchPage.css";
 
 interface SearchPageProps {
   path?: string;
-  community?: string;
 }
 
 // ES caps deep pagination at 10k; when the true count exceeds that, the
@@ -48,8 +47,8 @@ function formatResultsSummary(
   );
 }
 
-export function SearchPage({ community: slug }: SearchPageProps) {
-  const community = slug ? findCommunity(slug) : undefined;
+export function SearchPage(_props: SearchPageProps) {
+  const community = useCommunity();
   if (!community) return <NotFoundPage />;
   return <SearchPageInner community={community} />;
 }
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,20 +1,32 @@
 import Router from "preact-router";
 import { AuthProvider } from "./auth/AuthContext";
+import { CommunityProvider } from "./community/CommunityContext";
 import { AppShell } from "./components/layout/AppShell";
 import { SearchPage } from "./pages/SearchPage";
 import { RecordDetailPage } from "./pages/RecordDetailPage";
 import { NotFoundPage } from "./pages/NotFoundPage";
+import { URL_CHANGE_EVENT } from "./services/navigation";
+
+// preact-router intercepts internal <a href="/..."> clicks and updates the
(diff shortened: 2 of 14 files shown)


src/pages/SearchPage.tsx
Adam-Hammo: Should `useSearch` and `useCorpusTotal` call `useCommunity` themselves? I feel like that might better reflect the encompassing nature of communities in the application, instead of threading them through, but might be missing an implementation detail.

jaybea: Good insight, I've updated. It ended up being a bit nuanced regarding null handling:

Route-level null handling stays in SearchPage (!community → NotFoundPage) so the hooks don't own page semantics, they'll return idle state defensively when community is null.

I'm open to other implementations of course!

Adam-Hammo (approved):
Looks good!

PR #34, opened 2026-04-30
Add RawEnhancement.source to search and detail pages

### LINKED ISSUE #32: Show provenance at the reference and linked data level
Until https://github.com/destiny-evidence/destiny-repository/issues/633 has been done, the evidence repository UI will show duplicate results. Generally, each duplicate is a copy of a reference from a different source. This is potentially confusing/misleading to the user.

To make provenance clear, we should:
- Display the coding institution somewhere in the search result card. This can be found from the `source` on the first `RawEnhancement` (though should be mapped to a more friendly display name).
  - This should be rolled back after https://github.com/destiny-evidence/destiny-repository/issues/633
- Display the coding institution on the detailed investigation page. This time, this can be found from the `source` on the `RawEnhancement` that is pointed to by the displayed `LinkedDataEnhancement`'s `derived_from` relationship.
  - This will allow us to show provenance after https://github.com/destiny-evidence/destiny-repository/issues/633, as each LDE can display its own institution

Adam-Hammo:
Closes #32 

Adds the coding institution to the search and detail page.

This disambiguates duplicate references, and also may be useful when multiple LDEs are displayable under the same reference. (Though, we'll absolutely need a nicer pattern than a hardcoded map if we want to support this long term, particularly for new communities. We have [codedBy](https://vocab.evidence-repository.org/019d9463-2780-7243-b4de-e547386f2a90/1.0/property/019d9464-37b6-772b-a3ca-cef1ac603fda) in the vocab, perhaps we should start populating that?)

Screenshot 2026-04-30 at 4 27 09 pm
image

### DIFF
--- a/src/components/search/ResultRow.css
+++ b/src/components/search/ResultRow.css
@@ -90,6 +90,13 @@
   color: var(--color-text-secondary);
 }
 
+.row-coder {
+  font-family: var(--font-mono);
+  font-size: 12px;
+  font-weight: 400;
+  color: var(--color-text-tertiary);
+}
+
 /* Right column: deliberately un-positioned so clicks on its non-interactive
    children (stat-badges, empty space) fall through to the row-link overlay. */
 .row-right {
--- a/src/components/search/ResultRow.css
+++ b/src/components/search/ResultRow.css
@@ -92,8 +92,7 @@
 
 .row-coder {
   font-family: var(--font-mono);
-  font-size: 12px;
-  font-weight: 400;
+  font-size: var(--font-size-xs);
   color: var(--color-text-tertiary);
 }
 
--- a/src/services/codingInstitution.ts
+++ b/src/services/codingInstitution.ts
@@ -1,7 +1,8 @@
 import type { Enhancement, OtherEnhancement, Reference } from "@/types/models";
 import { extractLatestEnhancement } from "@/services/referenceUtils";
 
-// Substring match: real source values are more convoluted (e.g. "esea-coder-v3").
+// Substring match: real source values are more convoluted
+// (e.g. "eef-eppi-review", "ad_hoc_ingestors.iiie_ingestor@1.0").
 const INSTITUTION_PATTERNS: ReadonlyArray<readonly [string, string]> = [
   ["eef", "EEF"],
   ["iiie", "IIIE"],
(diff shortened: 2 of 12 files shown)


src/components/search/ResultRow.css
```diff
@@ -90,6 +90,13 @@
   color: var(--color-text-secondary);
 }
 
+.row-coder {
+  font-family: var(--font-mono);
+  font-size: 12px;
```
danielbreves:
```suggestion
  font-size: var(--font-size-xs);
```

src/components/search/ResultRow.css
```diff
@@ -90,6 +90,13 @@
...
+  font-weight: 400;
```
danielbreves: We may want to extract this to a variable too, though worth checking if we even need a reset here.

src/services/codingInstitution.ts
```diff
@@ -0,0 +1,46 @@
...
+  ["wwhge", "WWHGE"],
+];
+
+export function resolveCodingInstitution(
+  source: string | null | undefined,
+): string | null {
+  if (!source) return null;
+  const lower = source.toLowerCase();
+  for (const [pattern, label] of INSTITUTION_PATTERNS) {
+    if (lower.includes(pattern)) return label;
```
danielbreves: Could we do something like this?

```suggestion
    if (new RegExp(`(^|[^a-z])${pattern}([^a-z]|$)`).test(lower)) {
      return label;
    }
```

tests/services/codingInstitution.test.ts
```diff
@@ -0,0 +1,171 @@
...
+    robot_version: null,
+    derived_from: null,
+    created_at: createdAt,
+    content: {
+      enhancement_type: "raw",
+    },
+  };
+}
+
+function ldeEnh(
```
danielbreves: Could we use https://github.com/destiny-evidence/evidence-repo-ui/blob/e763b9aca248937a69ad35089a9f321d50231861/tests/fixtures.ts for these?

tests/services/codingInstitution.test.ts
```diff
@@ -70,6 +23,12 @@ describe("resolveCodingInstitution", () => {
     expect(resolveCodingInstitution("robot")).toBeNull();
   });
 
+  test("does not match patterns embedded in larger words", () => {
+    // "feeder" should not match "eef"; "messa" should not match "essa"; etc.
+    expect(resolveCodingInstitution("feeder")).toBeNull();
```
danielbreves: "beef" perhaps?

danielbreves (approved):

tests/services/codingInstitution.test.ts
Adam-Hammo: Choice!

PR #35, opened 2026-05-01
Add effect estimate and arm data cards (#10)

### LINKED ISSUE #10: Record detail page: Effect estimate and arm data cards

danielbreves:
Closes #10.

## Summary

Renders an EFFECT ESTIMATE inset inside each finding's outcome section, plus an ARM DATA table populated from `derivedFrom`. Layout and typography follow the [klim-b prototype](https://taxonomy-prototype.pages.dev/demo/investigation-klim-b) and the design-spec comment on #10.

Screenshot 2026-05-01 at 6 24 23 PM

## Highlights

- `classifyEstimate` decides pos/neg/ns from CI bounds (sign, exclusive of zero); exported and unit-tested separately so other surfaces can reuse it.
- `CIVisualization` places track, zero line, bar, dot, and tick labels in one shared percentage coord system inside a 20px-inset inner box, so axis-aligned elements stay pixel-aligned.
- Parser tolerates both `"_:armI"` and `{"@id": "_:armI"}` for `forCondition` / `derivedFrom`; arm/condition resolution is deferred to the renderer where the finding's own intervention/control IDs are available.
- `ArmDataTable` hides columns no arm populates, and falls back to rendering all finding arms when `derivedFrom` is absent.
- Adjustment tag follows the klim-b `.pill` ghost style; non-significant CI tint is the prototype's warm `rgba(154, 74, 18, …)`, kept component-local rather than promoted to a global token.
- `EffectEstimateData` only carries fields with a current consumer. `variance` and `forOutcome` were intentionally left unsurfaced.

## Test plan

- [x] `npm run typecheck`, `npm test` (298 pass; 26 new)
- [x] Visual smoke against seeded `…0010` record covers pos / neg / ns classifiers, sparse vs. full arm columns, present / absent `confidenceLevel` and `estimateSource`

### DIFF
--- a/src/components/EffectEstimateCard.css
+++ b/src/components/EffectEstimateCard.css
@@ -1,0 +1,100 @@
+.ee-card {
+  background: var(--color-bg-inset);
+  border: 1px solid var(--color-border);
+  border-radius: 10px;
+  padding: 22px 26px;
+  margin: 16px 0 12px;
+}
+
+.ee-card__label {
+  font-family: var(--font-mono);
+  font-size: 10px;
+  font-weight: 600;
+  text-transform: uppercase;
+  letter-spacing: 0.1em;
+  color: var(--color-text-tertiary);
+  margin: 0 0 14px;
+}
+
+.ee-card__headline {
+  display: flex;
+  align-items: baseline;
+  flex-wrap: wrap;
+  gap: 14px;
+  margin-bottom: 6px;
+}
+
+.ee-card__estimate {
+  font-family: var(--font-mono);
+  font-size: 30px;
+  font-weight: 700;
+  line-height: 1;
+  letter-spacing: -0.04em;
+}
+
+.ee-card__estimate--pos {
+  color: var(--color-text-primary);
+}
+.ee-card__estimate--neg {
+  color: #6b2020;
+}
+.ee-card__estimate--ns {
+  color: var(--color-text-secondary);
+}
+
+.ee-card__metric {
+  font-family: var(--font-mono);
+  font-size: 13px;
+  font-weight: 500;
+  color: var(--color-text-secondary);
+}
+
+.ee-card__se {
+  font-family: var(--font-mono);
+  font-size: 13px;
+  color: var(--color-text-tertiary);
+}
+
+.ee-card__ci {
+  font-family: var(--font-mono);
+  font-size: 14px;
+  color: var(--color-text-secondary);
+  margin-bottom: 4px;
+}
+
+.ee-card__ci-bracket {
+  color: var(--color-text-tertiary);
+}
+
+.ee-card__source {
+  font-size: 12px;
+  color: var(--color-text-tertiary);
+  font-style: italic;
+  margin: var(--spacing-xs) 0 0;
+}
+
+.ee-card__tags {
+  display: flex;
+  flex-wrap: wrap;
+  gap: var(--spacing-xs);
+  margin-top: var(--spacing-sm);
+}
+
+/* Ghost-outlined adjustment indicator (matches the prototype's .pill styling
+   in the klim-b override: transparent fill, strong border). */
+.ee-card__tag {
+  font-family: var(--font-body);
+  font-size: 12px;
+  font-weight: 500;
+  background: transparent;
+  border: 1px solid var(--color-border-strong);
+  border-radius: 3px;
+  padding: 3px 8px;
+  color: var(--color-text-secondary);
+}
+
+.ee-card__not-clustering {
+  font-size: 12px;
+  color: var(--color-text-tertiary);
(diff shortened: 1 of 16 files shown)


src/components/EffectEstimateCard.css
```diff
@@ -0,0 +1,100 @@
...
+  font-weight: 700;
+  line-height: 1;
+  letter-spacing: -0.04em;
+}
+
+.ee-card__estimate--pos {
+  color: var(--color-text-primary);
+}
+.ee-card__estimate--neg {
+  color: #6b2020;
```
Adam-Hammo: Nit, might be worth revisiting later - I personally found this colour quite hard to differentiate to the normal primary colour.

Adam-Hammo: We could/should also make this a variable, in line with our other colours?

src/components/EffectEstimateCard.css
```diff
@@ -0,0 +1,100 @@
...
+.ee-card__estimate {
+  font-family: var(--font-mono);
+  font-size: 30px;
+  font-weight: 700;
+  line-height: 1;
+  letter-spacing: -0.04em;
+}
+
+.ee-card__estimate--pos {
+  color: var(--color-text-primary);
```
Adam-Hammo: Should this be `#424A53`?

src/components/EffectEstimateCard.tsx
```diff
@@ -0,0 +1,114 @@
...
+      />
+
+      {showAdjustmentRow && (
+        <div class="ee-card__tags">
+          <span class="ee-card__tag">Baseline adjusted</span>
+        </div>
+      )}
+
+      {showNotClustering && (
+        <p class="ee-card__not-clustering">Not clustering adjusted</p>
```
Adam-Hammo: Nit, which I might be wrong on because I don't have context, but I feel like this sounds kind of clunky - maybe `"Not adjusted for clustering"`? And similar with baseline. I know this was what was specified in the issue though

Adam-Hammo: It also doesn't show as a card for me despite the classname, which is jarring below the baseline card.

Adam-Hammo: Maybe the tension is that baseline is shown if it is adjusted, and clustering is shown if it isn't? Idk, sorry for no clear direction!

src/components/EffectEstimateCard.tsx
```diff
@@ -0,0 +1,114 @@
...
+interface EffectEstimateCardProps {
+  estimate: EffectEstimateData;
+  arms: ArmData[];
+  intervention: InterventionData | null;
+  control: ControlConditionData | null;
+}
+
+export function classifyEstimate(estimate: EffectEstimateData): Significance {
+  const { ciLower, ciUpper } = estimate;
+  if (typeof ciLower !== "number" || typeof ciUpper !== "number") return "ns";
```
Adam-Hammo, line 20: Would it make sense here to use the pos/neg styles? (I don't know the answer)

Adam-Hammo (commented):
Looks fantastic, just a few small comments that may or may not require action!

src/components/EffectEstimateCard.tsx
danielbreves: Yea, this matches the prototype, my reading was that "Not clustering adjusted" was meant to be more like a footnote than a tag, but I agree "Not adjusted for clustering" reads better.

Screenshot 2026-05-05 at 11 54 53 AM

src/components/EffectEstimateCard.tsx
danielbreves: @Adam-Hammo how do you mean? My understanding is that when CI is missing we don't know whether the estimate is significant, so it could be misleading to display it as positive or negative.

src/components/EffectEstimateCard.css
danielbreves: I made it a dark blue now, what do you think?

Screenshot 2026-05-05 at 12 25 50 PM

Adam-Hammo (approved):
Looks great!

Issue #45, opened 2026-05-08
Export button and download flow on the search page


jack-o-the-green:
## Context

Allow the triggering of export creation from the search page. User clicks "Export to Excel" on a search result page, the SPA sends a request to destiny-repository `/download/` endpoint with the search query, receives a jsonl file of references in response. The frontend will then convert that jsonl file to the excel export (the conversion is a previous ticket) and download the xlsx fiel for the user. 

* Add an "Export to Excel" button to the meta bar next to the sort dropdown. Disabled when the result set is empty or when a previous export is still in flight. Disabled when total results exceed 10,000 (the destiny-repository cap). Prevent users from triggering an export we know we can't process
* On click, send the current search params (the same shape the page already builds for `GET /references/search/`) to destiny repositoy
   * Poll for completion. 
   * While the request is in flight, there should be an indication to the user that an export is being prepared
* If the export errors, a user-friendly error should be shown to the user. 
* When receiving a download url, trigger the converstion from jsonl to excel and download the file when it's ready.

jack-o-the-green:
Excel export implemented here https://github.com/destiny-evidence/evidence-repo-ui/blob/main/src/services/export/export.ts#L44

Will need to pass an acceptable vocabulary/context to the exporter. For now it probably makes sense to use whatever the most recent version of ESEA is, or to hardcode 1.1

Issue #47, opened 2026-05-10
Generalised excel exports


jack-o-the-green:
## Context 
The initial implementation of https://github.com/futureevidence/excel-export-experiments hard-codes the column names of the generated spreadsheet to the ESEA taxonomy, which was necessary for a first pass. This export will have been added to evidence-repo-ui as part of https://github.com/destiny-evidence/taxonomy-builder/issues/206 as an Azure function that works only for Jacobs data. 

We want to generalize this solution to support multiple vocabularies and move it out of it's custom Azure function implementation and into the destiny-repository core. 

## Requirements
* Spreadsheet structure is generated for a given published vocabulary
    * We'll need to solve how we export enhancements from previous versions of the same vocabulary. Might need to allow the user to specify. 
* Exports are limited to a single vocabulary at a time to ensure consistent spreadsheet generation 
* Implement a /download/ endpoint on destiny repository and a reference export service to handle the logic 
* evidence-repo-ui uses new download endpoint and azure function infrastructure cleared up. 

## Notes

PR #50, opened 2026-05-12
Jsonl -> excel spreadsheet conversion using SheetJs

jack-o-the-green:
## Summary

Adds an Excel export pipeline at `src/services/export/`. The
`exportReferencesToExcel(jsonlUrl, vocabularyUrl, contextUrl, filename)`
entry point streams a signed-URL JSONL of references through a
SheetJS workbook generator and triggers a browser download.

Whilst initially implemented in https://github.com/futureevidence/excel-export-experiments the export pipeline now reuses code from the evidence repository ui, including cached vocabularies from `vocabularyService` and reference utility functions from `referenceUtils` etc. 

Note that this does not include the UI changes for enabling exports, they will be implemented in the follow on issue #45  

## Changes to pre-existing files

- **`src/services/referenceUtils.ts`**
  - `extractLatestEnhancement` now applies a canonical-first dedup
    rule (enhancements with `reference_id === reference.id` win over
    duplicates even when a duplicate is newer; within the chosen
    bucket, latest `created_at` wins). Matches the projection logic
    on the destiny-repository side and makes display data consistent
    across record detail / search-result rows / export when a
    reference carries duplicate enhancements.
  - New `extractOpenAlexId` helper mirroring `extractDoi`.


src/services/export/generate.ts
```diff
@@ -0,0 +1,182 @@
...
+ */
+function appendSheet<R extends AnyRow>(
+  wb: XLSX.WorkBook,
+  name: string,
+  headers: ReadonlyArray<keyof R & string>,
+  rows: R[],
+): void {
+  const ws = XLSX.utils.aoa_to_sheet(rowsToAoa(headers, rows));
+  ws["!cols"] = colWidths(headers, rows);
+  (ws as unknown as { "!freeze"?: unknown })["!freeze"] = { xSplit: 0, ySplit: 1 };
```
Adam-Hammo: I'm not able to validate this in Excel, but this didn't freeze for me in Numbers or LibreOffice. An [old comment](https://github.com/SheetJS/sheetjs/issues/785#issuecomment-390501860) suggest it's a pro feature? Maybe we can remove if not needed.

src/services/export/buildRows.ts
```diff
@@ -0,0 +1,478 @@
...
+  ] as const satisfies ReadonlyArray<keyof InvestigationRow>,
+  arms: [
+    "reference_id",
+    "arm_id",
+    "intervention_name",
+    "intervention_description",
+    "control_description",
+    "intervention_duration_value",
+    "intervention_duration_supportingText",
+    "intervention_educationTheme",
```
Adam-Hammo, line 52: Just confirming we're happy hardcoding ESEA-themed columns in this pass and it's not a regression?

src/services/export/buildRows.ts
```diff
@@ -0,0 +1,478 @@
...
+    "source",
+    "title",
+    "authors",
+    "publication_year",
+    "doi",
+    "openalex_id",
+    "documentType",
+    "studyDesign",
+    "vocabulary",
+  ] as const satisfies ReadonlyArray<keyof InvestigationRow>,
```
Adam-Hammo, line 43: This is really nice

src/services/export/buildRows.ts
```diff
@@ -0,0 +1,478 @@
...
+    }
+    return (cv["@value"] ?? null) as CellValue;
+  }
+  return null;
+}
+
+/**
+ * Return the `@value` of a numeric/string coding annotation, or null.
+ * Numeric values pass through `round5` so the result matches the Python
+ * port's rounding behaviour.
```
Adam-Hammo: Nitpick on some comments - the references to Python may end up being confusing down the track when we leave the old script in the rear-view mirror

src/services/export/buildRows.ts
Adam-Hammo: (I can see that the effort required would be huge)

src/services/export/export.ts
```diff
@@ -0,0 +1,57 @@
...
+function triggerDownload(data: ArrayBuffer, filename: string): void {
+  const blob = new Blob([data], { type: XLSX_MIME });
+  const url = URL.createObjectURL(blob);
+  const anchor = document.createElement("a");
+  anchor.href = url;
+  anchor.download = filename;
+  document.body.appendChild(anchor);
+  anchor.click();
+  anchor.remove();
+  URL.revokeObjectURL(url);
```
Adam-Hammo:
```suggestion
  setTimeout(() => URL.revokeObjectURL(url), 0);
```
Safety nit to ensure the file is read before it's revoked

Adam-Hammo (commented):
Nothing major, looking good!

src/services/export/generate.ts
jack-o-the-green: Ahh good catch! I verified it in excel too - no freezing.

jack-o-the-green: Confirmed it doesn't work in excel too - good catch!

src/services/export/buildRows.ts
jack-o-the-green: Yep we're happy with the hardcoding in this pass, follow on issue is #47

src/services/export/buildRows.ts
jack-o-the-green: They will definitely be confusing down the track, thanks for catching!

Adam-Hammo (approved):
Looks good!

PR #55, opened 2026-05-18
Search export button and download flow

### LINKED ISSUE #45: Export button and download flow on the search page

danielbreves:
Closes https://github.com/destiny-evidence/evidence-repo-ui/issues/45.

Wires the search page to the queued-export endpoint from https://github.com/destiny-evidence/destiny-repository/pull/678. New `useSearchExport` hook does POST + 2s polling + JSONL→xlsx via the existing pipeline. Button hard-disabled on empty query, empty results, `is_lower_bound && count >= 10000`, or in-flight.

Vocab/context URLs come from `VITE_EXPORT_VOCABULARY_URL` / `VITE_EXPORT_CONTEXT_URL` — single-vocab-per-export per https://github.com/destiny-evidence/evidence-repo-ui/issues/47.

Extracts a small `Tooltip` component.

Browser SAS downloads against prod blob storage are blocked by CORS — tracked in https://github.com/destiny-evidence/destiny-repository/issues/688.

### DIFF
--- a/src/hooks/useSearchExport.ts
+++ b/src/hooks/useSearchExport.ts
@@ -30,17 +30,27 @@ export interface UseSearchExportResult {
 
 const POLL_INTERVAL_MS = 2000;
 
+/**
+ * Drives the export request → polling → download pipeline.
+ *
+ * **Cancellation.** Each call to `start` increments `runIdRef` and captures the
+ * new id in a closure. Every async callback compares its captured id against
+ * the live ref via `isCurrentRun()` and bails if they differ — meaning the
+ * user called `start` again, `reset`, or unmounted. State writes on stale
+ * callbacks would either warn (post-unmount) or corrupt the current run's
+ * state machine. AbortController would be the canonical alternative but
+ * would mean threading `signal` through the API client, the export module
+ * and `streamJsonlFromUrl`; revisit if any of those grow other reasons to
+ * abort.
+ */
 export function useSearchExport(): UseSearchExportResult {
   const [status, setStatus] = useState<ExportStatus>("idle");
   const [errorMessage, setErrorMessage] = useState<string | null>(null);
 
-  // Generation counter — bumped on reset/unmount. Callbacks that started under
-  // generation N bail if cancelRef.current !== N, so a late poll response from
-  // a cancelled run never writes state.
-  const cancelRef = useRef(0);
+  const runIdRef = useRef(0);
   const timeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
 
-  const clearTimeoutRef = useCallback(() => {
+  const clearScheduledPoll = useCallback(() => {
     if (timeoutRef.current !== null) {
       clearTimeout(timeoutRef.current);
       timeoutRef.current = null;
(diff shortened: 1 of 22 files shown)


src/hooks/useSearchExport.ts
```diff
@@ -0,0 +1,172 @@
...
+      runIdRef.current += 1;
+      const runId = runIdRef.current;
+      clearScheduledPoll();
+      setErrorMessage(null);
+
+      // The workbook falls back to raw CURIEs when Vocab/context URLs are missing.
+      // Warn so misconfiguration is visible in devtools without surfacing a user-facing error.
+      if (!EXPORT_VOCABULARY_URL || !EXPORT_CONTEXT_URL) {
+        console.warn(
+          "Export vocab/context URLs not configured; concept cells will contain raw CURIEs.",
```
jack-o-the-green, line 79: Do we want to do this here, or do we want to enforce the configuration of those URLs at startup? Happy either way.

src/pages/SearchPage.tsx
```diff
@@ -107,6 +138,52 @@ function SearchPageInner({ community }: { community: Community }) {
     navigate(buildSearchUrl(community.slug, { ...params, ...committed, sort, page: 1 }));
   }
 
+  const queryEmpty = params.q.trim() === "";
+  const hasYearFilter = params.startYear !== undefined || params.endYear !== undefined;
+  const noFilters = queryEmpty && !hasYearFilter;
```
jack-o-the-green: Is there value in disabling export for the `noFilters` case? Most cases will the caught by the 10,000 result cap, but I don't think it's a problem for anyone to export all references in a community if there's less than 10,000 of them.

src/pages/SearchPage.tsx
```diff
@@ -107,6 +138,52 @@ function SearchPageInner({ community }: { community: Community }) {
...
+    if (results.loading) return undefined;
+    if (overCap) {
+      return `Refine your search — exports are limited to ${EXPORT_MAX_RESULTS.toLocaleString()} results.`;
+    }
+    if (!hasResults) return "No results to export.";
+    return undefined;
+  })();
+  const exportAnnouncement = exportAnnouncementFor(exportJob.status);
+
+  function handleExport() {
```
jack-o-the-green, line 171: I feel it would be nicer to put the query construction `searchParams.ts` where`toQueryString` is implemented. Could do a `toExportSearchQuery`.

src/pages/SearchPage.tsx
danielbreves: I mostly wanted to prevent accidental or superfluous exports of all the data. If they want to export all the data they still can by entering * in the search box, but I recognise that's not intuitive or easy to discover. Happy to remove the restriction if we feel it's overly cautious though.

src/hooks/useSearchExport.ts
danielbreves: I thought since it's only used for exports and we can still export without a vocabulary/context it could be a soft warn rather than enforcement, but also happy either way.

jack-o-the-green: I think we're probably going to use them for faceted search as well - need something to build the filters off. But can make those changes later!

src/pages/SearchPage.tsx
jack-o-the-green: I think lets remove it for the time being.

src/hooks/useSearchExport.ts
danielbreves: Right, OK. I might leave it here for now, just because I'm not sure where's the best place to put it and it should be simple enough to move it once we need it for something else?

jack-o-the-green: Yeah that works!

jack-o-the-green (approved):
LGTM

PR #59, opened 2026-05-19
Display abstract on record detail page (closes #37)

### LINKED ISSUE #37: Record detail page: Display Abstract
### Summary
Where data is available, display the Abstract on the Record Detail Page

jaybea:
## Summary

- Render the abstract of a reference on the record detail page when an `AbstractContentEnhancement` is present. Heading + body sit between the InvestigationCard and the findings block; absent / empty / whitespace-only / `:unav`-sentinel abstracts render nothing.
- Unify abstract selection across the codebase via a new `extractAbstract` utility: canonical-first, longest-wins, newest tie-break. Replaces `ResultRow`'s inline `findAbstract()` which used `.find()` and would return whichever abstract appeared first in array order. Regression case: W4411634320, which carries a newer truncated re-ingest alongside the older intact 1896-char body.
- HTML entities decoded via a small `decodeHtmlEntities` utility built on `<textarea>.innerHTML` (decodes named, numeric, and hex entities; preserves tags as literal text). The textarea content model guards against the silent tag stripping that `DOMParser` would do. Mojibake from the EEF EPPI parser is rendered as-is; the fix belongs upstream.
- Strip the literal section label "Abstract" when the source data captures it as the first word of the body (~1.5% of the local corpus). Shared `stripAbstractLabelPrefix` helper applied in both `AbstractSection` (detail page) and `ResultRow` (search snippet).

## Validation performed

- Typecheck clean (`npm run typecheck`).
- Full test suite: 430 passing (baseline 405 + 25 new). 6 commits, each a green state.
- New tests: 5 `decodeHtmlEntities`, 6 `stripAbstractLabelPrefix` (table-driven), 7 `extractAbstract` (including canonical-vs-duplicate bucket coverage), 2 `ResultRow` regressions (longest-wins + prefix strip), 7 `AbstractSection` policy + prefix strip, 2 `RecordDetailPage` wiring.
- Production build clean (`npm run build`).
- Live verification against local destiny-repository (14,357 abstracts indexed, ESEA community):
  - Abstract heading and body render below the InvestigationCard on OpenAlex-sourced refs.
  - Entity decoding confirmed: `p &gt; 0.05` renders as `p > 0.05` on `019dc685-59f3-77de-b81a-4cf7cf1148e6`.
  - Tag-passthrough confirmed: `<em>r</em>` renders as literal text on `019d986a-970a-75aa-9cbe-f0c618adb0a8` (would have been stripped by DOMParser).
  - Mojibake passes through unchanged on EEF-sourced refs (`â€` artifacts visible in body, matching the search-row rendering above).
  - Leading "Abstract" label removed in both detail page and search-row snippet (confirmed on `019d9868-f3ba-7619-9980-b459ee80549c`, an EEF-ingested ref).
  - Search-row snippets still render. The longest-wins switch in `ResultRow` only changes output for references with two or more abstracts of differing lengths; for the typical single-abstract reference (the vast majority of the corpus) the row looks identical to before, so live verification confirms the swap is non-breaking rather than directly observing the regression fix.
  - No new console errors or warnings introduced.

## Implementation notes

- `extractAbstract` deliberately diverges from `extractLatestEnhancement`'s newest-wins selection (length wins, recency breaks ties) to rescue the W4411634320 truncation case. Canonical-vs-duplicate bucketing is preserved so a duplicate's longer body can't override the canonical record. Rationale documented inline at `src/services/referenceUtils.ts`.
- `decodeHtmlEntities` uses `<textarea>.innerHTML` over `DOMParser` so tags inside abstract text aren't silently stripped. Live verification surfaced a real instance of this (`<em>r</em> = 0.7`) on a ref from the local corpus.
- `stripAbstractLabelPrefix` is case-sensitive (only the literal label `Abstract`, not `abstract`/`ABSTRACT`). Live verification confirmed no false positives across the search-result sample.
- `AbstractSection` CSS uses pre-existing tokens (`--spacing-lg`, `--spacing-md`, `--font-size-body`, `--color-text-primary`); `unicode-bidi: plaintext` future-proofs against mixed-direction abstracts; `white-space: pre-line` future-proofs against synthetic abstracts with paragraph breaks.

### DIFF
--- a/src/components/AbstractSection.css
+++ b/src/components/AbstractSection.css
@@ -1,0 +1,17 @@
+.abstract-section {
+  margin: var(--spacing-lg) 0;
+}
+
+.abstract-section__heading {
+  font-size: 18px;
+  font-weight: 600;
+  margin: 0 0 var(--spacing-md);
+}
+
+.abstract-section__body {
+  font-size: var(--font-size-body);
+  line-height: 1.55;
+  color: var(--color-text-primary);
+  white-space: pre-line;
+  unicode-bidi: plaintext;
+}
--- a/src/components/AbstractSection.css
+++ b/src/components/AbstractSection.css
@@ -1,11 +1,9 @@
 .abstract-section {
-  margin: var(--spacing-lg) 0;
+  margin: var(--spacing-lg) 0 0;
 }
 
 .abstract-section__heading {
-  font-size: 18px;
-  font-weight: 600;
-  margin: 0 0 var(--spacing-md);
+  margin-top: 0;
 }
 
 .abstract-section__body {
--- a/src/pages/RecordDetailPage.tsx
+++ b/src/pages/RecordDetailPage.tsx
@@ -1,6 +1,7 @@
 import { useMemo } from "preact/hooks";
 import { useCommunity } from "@/community/CommunityContext";
 import {
+  extractAbstract,
   extractBibliographic,
   extractLinkedData,
   extractLinkedDataEnhancement,
@@ -14,6 +15,7 @@ import {
 import { useReference } from "@/hooks/useReference";
 import { useVocabulary } from "@/hooks/useVocabulary";
 import { useContextPrefixes } from "@/hooks/useContextPrefixes";
+import { AbstractSection } from "@/components/AbstractSection";
 import { InvestigationCard } from "@/components/InvestigationCard";
 import { FindingsSection } from "@/components/FindingsSection";
 import { NotFoundPage } from "./NotFoundPage";
@@ -95,6 +97,7 @@ export function RecordDetailPage({ id }: RecordDetailPageProps) {
   const codingInstitution = lde
     ? extractLinkedDataCodingInstitution(reference, lde)
     : null;
+  const abstract = extractAbstract(reference);
 
   return (
     <div class="record-detail-page">
@@ -112,6 +115,7 @@ export function RecordDetailPage({ id }: RecordDetailPageProps) {
           hasInvestigation={hasLinkedData}
           vocabUnavailable={!!vocabUnavailable}
         />
+        <AbstractSection abstract={abstract} />
         {investigation && investigation.findings.length > 0 && (
           <>
             {vocabUnavailable && (
--- a/src/pages/RecordDetailPage.tsx
+++ b/src/pages/RecordDetailPage.tsx
@@ -15,7 +15,6 @@ import {
 import { useReference } from "@/hooks/useReference";
 import { useVocabulary } from "@/hooks/useVocabulary";
 import { useContextPrefixes } from "@/hooks/useContextPrefixes";
-import { AbstractSection } from "@/components/AbstractSection";
 import { InvestigationCard } from "@/components/InvestigationCard";
 import { FindingsSection } from "@/components/FindingsSection";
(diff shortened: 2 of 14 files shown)


tests/components/search/ResultRow.test.tsx
```diff
@@ -282,4 +283,81 @@ describe("ResultRow", () => {
...
+    });
+    render(<ResultRow communitySlug="esea" reference={ref} />);
+    expect(
+      screen.getByText("This paper aims to investigate."),
+    ).toBeInTheDocument();
+    expect(screen.queryByText(/^Abstract This/)).toBeNull();
+  });
+
+  test("selects the longer abstract when two abstract enhancements differ in length (longest-wins via extractAbstract)", () => {
+    const ref: Reference = {
```
danielbreves: Could we use fixtures.ts for this?

src/pages/RecordDetailPage.tsx
```diff
@@ -112,6 +115,7 @@ export function RecordDetailPage({ id }: RecordDetailPageProps) {
           hasInvestigation={hasLinkedData}
           vocabUnavailable={!!vocabUnavailable}
         />
+        <AbstractSection abstract={abstract} />
```
danielbreves: It looks a bit odd outside a card. Could we put it within the `InvestigationCard` after the DOI?

Image

src/pages/RecordDetailPage.tsx
```diff
@@ -95,6 +97,7 @@ export function RecordDetailPage({ id }: RecordDetailPageProps) {
   const codingInstitution = lde
     ? extractLinkedDataCodingInstitution(reference, lde)
     : null;
+  const abstract = extractAbstract(reference);
```
danielbreves, line 99: Should we `stripAbstractLabelPrefix` and `decodeHtmlEntities` here too?

danielbreves:
Could we make some of the inline comments a bit more succinct?

src/components/AbstractSection.css
```diff
@@ -0,0 +1,17 @@
+.abstract-section {
+  margin: var(--spacing-lg) 0;
+}
+
+.abstract-section__heading {
+  font-size: 18px;
+  font-weight: 600;
```
danielbreves: Could we use vars for these?

tests/components/search/ResultRow.test.tsx
jaybea: Yes, done!

src/pages/RecordDetailPage.tsx
jaybea: Yes, good call, looks better, uses existing mono-caps treatment for the ABSTRACT section names...

image

src/pages/RecordDetailPage.tsx
jaybea: Yes, I centralised this so it gets applied everywhere.

src/components/AbstractSection.css
jaybea: Yes, reused the shared `lg-section-label` utility class rather than redeclaring tokens inline.

jaybea:
@danielbreves  I think I addressed all comment, let me know if it needs further tweaking!

danielbreves (approved):
LGTM!

Issue #62, opened 2026-05-20
Add ability to search by linked_data_concept to search and export queries.


jack-o-the-green:
## Background 

Step one of supporting faceted search to allow us to make requests from the evidence-repo-ui to destiny-repository that contain search facets. 

This is a small ticket to plumb this in

## Technical Scope

* Add a `searchFacets: string[]` field to `SearchParams` so callers can attach Lucene fragments. Each facet will contain one or more `linked_data_concepts:"URI"` joined by ORs. e.g. `linked_data_concepts:"URI_A" OR linked_data_concepts:"URI_B"`.
    * If it's easy to add some checks at construction time, do that. 
* AND-join searchFacets into the API call's `q` parameter.
  * Suggest wrapping each search facet in parentheses 
* An empty base `q` with non-empty search facets should join like `* AND (facet1) AND (facet2)` 
* `parseSearchParams` should split **trailing** ` AND (<facet>)` segments out of the raw `q` into `searchFacets`. 
   * Constrain to fragments that start with `linked_data_concepts:` if possible.

Expect the majority of changes to be inside `searchParams.ts`, `apiClient.ts` and `useSearch.ts`

### Out of scope (no change)

- `src/hooks/useSearchDraft.ts`: facets won't be edited from the search bar input, but from the FilterDrawer which will be implemented in subsequent tickets.

PR #69, opened 2026-05-21
Require vocabulary url and derive concept hierarchy in cached data

### LINKED ISSUE #63: Make vocabulary concept schemes available for faceted search filters
## Context

Filters need to be built off a specific vocabulary, using the concept schemes to construct the filters. 

At a first pass we are happy to support only the latest version of the ESEA vocabulary. We're currently doing this for exports where the vocabulary is configured on the repository using environment variables. https://github.com/destiny-evidence/evidence-repo-ui/pull/64

We want to move that export vocabulary version pinning into the Community, and use it for both exports and filter generation.

This gets it's own ticket because it also changes how exports access their vocabulary, so a small change now is nicer rather than bundling it into the FilterDrawer implementation later. 

## Technical Scope

* Communities in communities.ts should be updated to include vocabulary and context urls (these could be set by environment variables, implementer decide). 
    * If env variables and they're missing, application should fail to start.
    * Add the correct variables for the esea community
* Exports should then use these vocabulary and context urls from the community to translate between concepts and preflabels etc. 
* We need to extract the vocabulary hierarchy for filtering - concept scheme -> broader concept -> narrower concepts. There is a buildVocabularyData function in `vocabulary.ts` which currently filters all skos concepts only. Explore whether it is appropriate to extend this into a hierarchical data format, or otherwise add a separate hierarchical data format that is also cached.

Adam-Hammo:
Closes #63 by extending `VocabularyData` to include the concept scheme hierarchy.


jack-o-the-green (approved):
Looks good to me. 

Have one thought about how this interacts with #62/#72. Do we want to pull the change you've described here for context compaction into https://github.com/destiny-evidence/evidence-repo-ui/pull/72#discussion_r3279399730 into this PR?

If we use context compaction, I'm assuming we'll _also_ start stripping not just the vocabulary base, but the `https://vocab.evidence-repository.org/` prefixes as well (ESEA contains both) whereas the current implementation only strips the configured vocab base. 

In the current implementation this will keep fully qualified uris in the concept schemes, so we'll need to strip the vocabulary base out of the concepts when constructing Lucene queries inside filters (I think this the correct way to do things). But I think how we do that is impacted by the above question. 

Also wondering if we should add some kind of note to https://github.com/destiny-evidence/destiny-repository/issues/695 about how to unpick this later?

jaybea (approved):
Very nice, I poked and prodded it and it came up trumps!

- Fail-fast contract works as intended. Initial run with `VITE_ESEA_*_URL` missing produced a clean throw pointing at `communities.ts:5` with the missing var name; no guessing needed to fix.
- Record detail page renders concept labels correctly (THEME breadcrumb "Teaching and Learning › Targeted Support", plus IMPLEMENTER and IMPLEMENTATION FIDELITY chips). Since `buildVocabularyData` builds labels/broader/definitions/schemes in a single pass, label rendering working confirms the whole function ran end-to-end including the new scheme-parsing path.

On the rebase with PR #72, and the addition of vocabBase, that was poorly named. Should have been something more like:

```ts
/** Base IRI for concept identifiers; mirrors the JSON-LD context `esea` prefix.
 *  Drop when destiny-evidence/destiny-repository#695 (structured
 *  `linked_data_concepts` query param) lands; read from `contextUrl` after that. */
conceptIriBase: string;
```

But happy to defer to what you think is best in the rebase.

PR #71, opened 2026-05-21
FilterFilterCard + ConceptSchemeFilter

### LINKED ISSUE #66: Implement ConceptSchemeFilter + FilterCard Components
## Background 
Proposal at #25 

MVP prototype: https://claude.ai/design/p/4df77d9c-8dec-45f9-9a19-65cbce3b7cab?file=MVP+prototype.html

This ticket is for implementing the FilterCard and ConceptSchemeFilter components described in the proposal. These will form the building blocks of the Filter Drawer implemented in the next ticket. 

## Technical Scope 

Refer to prototype for component visualization, styling does not need to match exactly.

* FilterCard
    * Add a FilterCard component that is responsible for the title of the filter, and whether the Filter inside it is minimized or expanded. 
    * Should also display if the filter active. A filter should be responsible for communicating how it is active i.e. the text to display here. 
    * We expect FilterCards to contain multiple types of filter later on, so only implement generalized functionality here. The FilterDrawer will eventually be responsible for creating FilterCards + Child Filters. 
* ConceptSchemeFilter
   * Add a ConceptSchemeFilter that is responsible for displaying a concept scheme from a taxonomy as drop down as well as generating it's own lucene query fragment. 
   * Should be build from the same data structure that is implemented in #63 
   * Needs to handle
       * It's own rendering 
       * Converting itself to a Lucene fragment 
   * Parent compoents must be able to (through functions on child)
       * Determine if the filter is empty (skip empty filters at fragment join time)
       * Display a summary of what is selected in the filter 
           * For concept scheme this will be an "X selected" displayed on FilterCard if the filter is not empty.  
       * Access the Lucene fragment

## Example usage
```
<FilterCard title={config.label}>
  <ConceptSchemeFilter state={state} ... />
</FilterCard>
```
```

### PR DESCRIPTION
**Pull request by jack-o-the-green**
# FilterCard + ConceptSchemeFilter

Closes #66.

Two building blocks the upcoming FilterDrawer ticket will assemble.

## What's in

- **`FilterCard`** — generic expand/collapse shell that owns its title and an optional "active state" summary shown while collapsed. Knows nothing about what filter sits inside it.
- **`ConceptSchemeFilter`** — controlled checkbox tree bound to one SKOS concept scheme. Renders flat (DocumentType) and hierarchical (Outcome) schemes alike; clicking a parent selects (or clears) the whole subtree in one go.
- **Pure state helpers** (`conceptSchemeFilterState.ts`) — `Concept` / `ConceptScheme` types, an opaque `ConceptSchemeFilterState`, and the helpers a parent needs to compose filters: `emptyConceptSchemeState`, `conceptSchemeStateFromUris`, `selectedUris`, `isEmpty`, `summary`, `toggleConcept`, `toggleConceptSubtree`, `toLuceneFragment`. State is parent-owned, so the future drawer can load selections straight from the URL.
- **`/dev/filters` sandbox** (DEV-only route) — exercises the components against a real ESEA OutcomeScheme slice with a live state echo. Gated by `import.meta.env.DEV`; JS tree-shakes from production, a tiny amount of CSS leaks.

### CODE REVIEW

File: src/components/filters/ConceptSchemeFilter.tsx [hunk 1]
```diff
@@ -0,0 +1,84 @@
...
+  onChange: (next: ConceptSchemeFilterState) => void;
+}
+
+function ConceptItem({ concept, state, onChange }: ConceptItemProps) {
+  const hasDefinition = !!concept.definition;
+  const labelClass = hasDefinition
+    ? "concept-scheme-filter__label concept-scheme-filter__label--has-tooltip"
+    : "concept-scheme-filter__label";
+  const labelNode = (
+    <span class={labelClass} tabIndex={hasDefinition ? 0 : undefined}>
```
Adam-Hammo, line 35: This opens on `tab`, which makes navigating with the keyboard a bit clunky as it goes to definition instead of next checkbox. Of course, we also would want an accessible way of opening the definition. Probably not a time sink to enter into now but just to be aware of.

src/components/filters/ConceptSchemeFilter.tsx
```diff
@@ -0,0 +1,84 @@
...
+          onChange={() => onChange(toggleConceptSubtree(state, concept))}
+        />
+        {hasDefinition ? (
+          <Tooltip text={concept.definition}>{labelNode}</Tooltip>
+        ) : (
+          labelNode
+        )}
+      </label>
+      {hasChildren && (
+        <ul class="concept-scheme-filter__children">
```
Adam-Hammo, line 56: Should the parent checkbox toggle on when all its children are checked, and vice-versa? Perhaps with [indeterminate](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:indeterminate#checkbox_radio_button) checkboxes too in the middle? 
I'm not sure what the actual expected UX is so feel free to reject me!

Adam-Hammo (approved):
Two notes, pre-approving for if we decide to do nothing as I'm off on Monday.

src/components/filters/ConceptSchemeFilter.tsx
mootpointer: I think toggling with parent is good, but if it's going to take much effort, I'd happily follow up with it.

jack-o-the-green: With current search implementation these would be two different searches/search result counts, here's the different behavior currently. 

* If all children but not parent are selected, we get leaf concepts only in the search and miss anything that's bucketed into the higher parent concept. 

* If parent + children selected then parent concept included in the result. 

If the parent concept toggles when all children are selected, we lose the ability to do child-concept-only search (because right now untoggling the parent untoggles children). 

However, I think this is probably the right way to go regardless, as it feels like users will assume that if all children are toggled, they're getting everything in that bucket (parent included). 

I'll see how quick the change is, but just wanted to highlight the loss of the child-concept-only search that would come from the change.

src/components/filters/ConceptSchemeFilter.tsx
jack-o-the-green: I will raise a ticket for UI accessibility improvements - there's some work on focus capture I deferred from the FilterDrawer implementation as well as it was more complexity

src/components/filters/ConceptSchemeFilter.tsx
jack-o-the-green: This was a pretty simple and isolated change to make, so have done so!

mootpointer: I think this makes sense, but we should probably make the hierarchy transitive in the future so this is explicitly encoded in the data, not just enforced in the UI.

Issue #67, opened 2026-05-20
Add FilterDrawer and plumb it all together


jack-o-the-green:
## Background 
Proposal at #25 

MVP prototype: https://claude.ai/design/p/4df77d9c-8dec-45f9-9a19-65cbce3b7cab?file=MVP+prototype.html

**Note**: Moving the year range into the FilterDrawer is not in scope for this ticket, it has its own follow-up

This ticket is for building our FilterDrawer component and tying our faceted search implementation together.

### FilterDrawer

* Implement FilterDrawer component (refer to prototype for visualization and behavior). This should have 
    * A scrollable section of FilterCards with ConceptSchemeFilters for each concept scheme in the ESEA taxonomy. Should be able to access these from the data structure in #63 
    * A sticky "Update Results" button that is greyed out if no changes have been made to the filters 
    * A "Cancel" button
    * A Reset button that removes all filters
        *  Should trigger navigation to the url with all facets removed but previous search terms (e.g. `q=phonics`) should be retained. 
    * When the drawer is active the user should not be able to click away from it to close. They need to interact with one of the buttons 

*  The FilterDrawer should maintain a draft of search facets, which are then applied to the searchParams/API call implemented in #62 when the "Update Results" button is clicked by the user. 
    * This matches the current pattern used by the year range search terms
    * On page load, if there are facets selected, the search draft in the drawer should be populated with those selected facets. 

### FilterDrawer Trigger 
* Add a "Refine" button as shown on the prototype that triggers the opening of the FilterDrawer
* When filters are applied this should display a number of how many filters 
    * To start with this should just be the count of non-empty concept scheme filters/the number of Lucene query fragments in the search parameters.
    * We can look at updating this to be something more comprehensive later. 

## Note
If any paticular UI behaviour is causing a rabbithole, raise it with the team and discuss necessity of implementing it in the first pass. Better to do lots of small follow ups than delay landing this one.

acdonald:
> A Close button

should be a "cancel' button

jack-o-the-green:
Fixed!

PR #72, opened 2026-05-21
Add linked_data_concepts faceted search (#62)

### LINKED ISSUE #62: Add ability to search by linked_data_concept to search and export queries.

jaybea:
Closes #62.

## Summary

Plumbs `searchFacets: string[]` through `SearchParams`, the URL, `useSearch`, and the export flow. The upcoming FilterDrawer (Issue #67) can now attach `linked_data_concepts:"<URI>"` clauses to live search and export requests. This PR is the data path only; no UI for editing facets.

URIs in typed state and the URL use a compact `Scheme/C-id` form (e.g. `EducationLevelScheme/C00002`). New helpers `expandFacets(facets, vocabBase)` (API boundary) and `compactFacets(facets, vocabBase)` (URL parse) move URIs between compact and fully-qualified form. `vocabBase` lives on `Community` alongside `defaultAnnotations`, with a drift guard asserting the esea community's value matches the JSON-LD context fixture's `esea` prefix.

The wire-format builder `buildFacetedQuery(q, facets)` joins facets as `(base) AND (facet1) AND (facet2)`, with `*` substituted for empty base. Non-empty base is wrapped in parens for boolean-precedence safety; without the wrap, `phonics OR reading AND (facet)` would bind in Lucene as `phonics OR (reading AND (facet))`.

`parseSearchParams` peels strict trailing `\s+AND\s+\(linked_data_concepts:[^)]+\)\s*$` clauses out of `q`, strips a lone `*` sentinel, and unwraps a single-level paren-wrap. Anything outside that strict grammar stays in `q` so user-typed Lucene is never silently dropped. `toQueryString(parseSearchParams(url))` is idempotent.

`SearchPage`'s `showMetaBar` and `showSummary` gates now include facets so a facet-only URL renders the meta region (regression-tested at unit and page layers, including the loading state).

`useSearchDraft` is unchanged; facets are not edited from the search bar.

## Stopgap note

destiny-repository#695 proposes a structured `?concept=URI_A&concept=URI_B` API parameter that would replace this approach. The team deferred #695 on 2026-05-21. When it lands, expect to remove the `parseSearchParams` peel/unwrap/compact logic, `buildFacetedQuery`, `expandFacets`, `compactFacets`, and the URL grammar shipped here.

URL verbosity (each selected concept emits a `linked_data_concepts:"..."` clause with field-name repetition) is the known cost. At 20 selections across 5 schemes the URL reaches roughly 1,500 to 2,000 encoded characters. Within browser limits, but visually rough; part of why #695 is the proper long-term shape.

## Migration note

The peel regex matches any trailing `\s+AND\s+\(linked_data_concepts:[^)]+\)\s*$`. Shareable URLs from before this PR that contain hand-typed `linked_data_concepts:` clauses will be reclassified into `searchFacets` on parse, with compact-looking values expanded via `vocabBase` at the API boundary. This is deliberate: the field's semantics are vocab URIs, so compact-looking literals never match the backend index anyway, and the wire query is preserved either way.

## Validation performed

- Full test suite green: 496 tests across 48 files.
- Typecheck and production build clean.
- Live dev server verified end-to-end against the local destiny-repository docker stack:
  - Compact URI facet URL: `q=phonics` plus one facet returned 63 results; the same facet alone returned 1,854; the GET network request contained the fully-qualified URI `https://vocab.esea.education/EducationLevelScheme/C00002`.
  - Facet-only URL (empty base) showed an empty search input and rendered the meta bar with the summary line.
  - Plain query `q=phonics` returned 105 results unchanged.
  - Browse mode unchanged.
  - Export POST request URL contained the fully-qualified wire query verbatim.

## Follow-ups (out of scope)

- FilterDrawer UI to edit `searchFacets` (Issue #67).
- Migration to destiny-repository#695's structured `?concept=` parameter once that lands backend-side.

### DIFF
--- a/src/services/searchParams.ts
+++ b/src/services/searchParams.ts
@@ -93,3 +93,21 @@ export function buildFacetedQuery(q: string, facets: string[]): string {
   const base = trimmed === "" ? "*" : `(${trimmed})`;
   return [base, ...facets.map((f) => `(${f})`)].join(" AND ");
 }
+
+const LDC_URI = /(linked_data_concepts:")([^"]+)(")/g;
+
+export function expandFacets(facets: string[], vocabBase: string): string[] {
+  return facets.map((f) =>
+    f.replace(LDC_URI, (_, prefix, uri, suffix) =>
+      uri.startsWith("http") ? `${prefix}${uri}${suffix}` : `${prefix}${vocabBase}${uri}${suffix}`,
+    ),
+  );
+}
+
+export function compactFacets(facets: string[], vocabBase: string): string[] {
+  return facets.map((f) =>
+    f.replace(LDC_URI, (_, prefix, uri, suffix) =>
+      uri.startsWith(vocabBase) ? `${prefix}${uri.slice(vocabBase.length)}${suffix}` : `${prefix}${uri}${suffix}`,
+    ),
+  );
+}
--- a/src/services/searchParams.ts
+++ b/src/services/searchParams.ts
@@ -15,6 +15,7 @@ export interface SearchParams {
   startYear: number | undefined;
   endYear: number | undefined;
   sort: SortOption | undefined;
+  searchFacets: string[];
 }
 
 // Unknown values fall back to undefined (relevance).
@@ -49,7 +50,7 @@ export function parseSearchParams(search: string): SearchParams {
 
   const sort = parseSort(params.get("sort"));
 
-  return { q, page, startYear, endYear, sort };
+  return { q, page, startYear, endYear, sort, searchFacets: [] };
 }
 
 export function toQueryString(params: SearchParams): string {
--- a/src/services/searchParams.ts
+++ b/src/services/searchParams.ts
@@ -85,7 +85,10 @@ export function parseSearchParams(search: string, vocabBase?: string): SearchPar
 
 export function toQueryString(params: SearchParams): string {
   const out = new URLSearchParams();
-  if (params.q) out.set("q", params.q);
+  const qSerialised = params.searchFacets.length > 0
+    ? buildFacetedQuery(params.q, params.searchFacets)
+    : params.q;
+  if (qSerialised) out.set("q", qSerialised);
   if (params.startYear !== undefined) out.set("start_year", String(params.startYear));
   if (params.endYear !== undefined) out.set("end_year", String(params.endYear));
   if (params.sort !== undefined) out.set("sort", params.sort);
--- a/src/services/searchParams.ts
+++ b/src/services/searchParams.ts
@@ -31,12 +31,42 @@ function parseDecimalInt(raw: string | null): number | undefined {
   return Number.isSafeInteger(n) ? n : undefined;
 }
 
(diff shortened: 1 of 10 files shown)


src/services/searchParams.ts
```diff
@@ -30,12 +31,42 @@ function parseDecimalInt(raw: string | null): number | undefined {
...
+
+export function parseSearchParams(search: string, vocabBase?: string): SearchParams {
   const params = new URLSearchParams(
     search.startsWith("?") ? search.slice(1) : search,
   );
 
-  const q = (params.get("q") ?? "").trim();
+  let q = (params.get("q") ?? "").trim();
+
+  const facetTail = /\s+AND\s+\(\s*(linked_data_concepts:[^)]+?)\s*\)\s*$/;
```
Adam-Hammo, line 56: The whole pattern of compacting URIs and re-extracting them via regex feels a little superfluous to me in light of https://github.com/destiny-evidence/destiny-repository/issues/695. For a temporary MVP, can we just get away with passing through URIs as-is?

We also have `context` per-community in https://github.com/destiny-evidence/evidence-repo-ui/pull/69 which would allow us to do the compaction and expansion dynamically without hardcoding `vocabBase` to `community`.

I don't feel strongly either way as it's transient code, just wondering what the shortest path to the final state is.

Adam-Hammo (approved):
One pondering, but not a blocker. I think we can just keep it in mind when we return after making linked data concepts an API parameter.

src/services/searchParams.ts
jaybea: Yeah, I was concerned about the total length of the url, given the MVP shows you could easily select a dozen or two facets. on vocabBase, #69 wasn't available yet, and I wanted a synchronous, always there value to avoid introducing loading state on search critical path. Maybe overkill for this V0, yeah.

Anyway, yes this all goes away with #695, a much better solution.

Issue #74, opened 2026-05-21
Move year range filter from SearchBar to FilterDrawer


jack-o-the-green:
## Context

Shifting the year range filter into the FilterDrawer was de-scoped from the initial deliverable of faceted search despite being in our MVP prototype as it required different handling of it's search parameters and wasn't required to make faceted search work.

MVP prototype: https://claude.ai/design/p/4df77d9c-8dec-45f9-9a19-65cbce3b7cab?file=MVP+prototype.html

This ticket is a follow on to move the year range filter in to the filter drawer.

## Technical Scope
Move the year range filter from the SearchBar into the top of the filter drawer. The following will be required
* Implemeting a `YearRangeFilter` that has the same or similar contract with the `FilterDrawer` as the `ConceptSchemeFilter` components. 
     * This should live inside a `FilterCard` component which will handle minimization, title, and summary dispaly
 * Inputs to the year range filter should update the searchParams as they do now, no search behavior is changed. 
     * All filters will be in the same draft search after this change, the origianl page will include the search query only
     * Do what clean up you can between different draft searches now.

Issue #73, opened 2026-05-21
Add facet counts to filter drawer


Adam-Hammo:
# Context 

We want to show the number of references in each facet, respective to the currently applied filters. This allows users to more easily explore the available data.

# Technical scope

Add the facet counts from https://github.com/destiny-evidence/destiny-repository/issues/698 to the filter drawer. This should be a near-identical API call to the search, this time to `/search/facets/?facet=concepts`. However, be wary of 

These should look as in the [prototype](https://claude.ai/design/p/4df77d9c-8dec-45f9-9a19-65cbce3b7cab?file=Fast+follow+prototype.html).

They should be loaded on page load, before the drawer is opened, based on the search query. 

Consider updating every time a filter is selected - this will need an assessment of query time and a consideration of UX (eg do we show a loading icon while that happens?). I'm not sure what the plumbing looks like, but the query building might be nasty, or it might be solvable straight away! If this snowballs, we should probably defer to a follow-up.

Per the limitation mentioned in https://github.com/destiny-evidence/destiny-repository/issues/698, siblings of selected filters should **not** show facet counts. This is not currently represented in the prototype.

PR #75, opened 2026-05-24
FilterDrawer basic implementation

jack-o-the-green:
Closes of #67 

Adds the FilterDrawer and allows application of concept scheme filtering to search results. Very quick implementation that can be demoed to our stakeholders. 

When reviewing the PR please focus on unintended behavior over beautiful UI goodness. There will be a follow up PR to improve the design. 

Best way to test this is to checkout the branch, connect your local to staging (or prod) destiny repo and have a go with the filter drawer. 

## Changes

- **FilterDrawer** — modal dialog (backdrop, escape-to-close,
  body scroll lock, focus restoration) that renders one `FilterCard` per
  ConceptScheme in the ESEA vocabulary. Apply is disabled until the draft differs from the applied
  facets; Reset clears the draft; Cancel reverts.
- **RefineButton** — search-bar trigger with a selected-count badge and a
  tooltip-on-disabled state for when vocabulary is loading or
  unavailable.
- **parseFacets** — reverses `toSearchFacet` across the whole
  `SearchParams.searchFacets` array, bucketing concept URIs back into
  per-scheme state. URIs not found in the current schemes are dropped
  defensively (stale URLs after a vocabulary change).
- **SearchPage wiring** — `useVocabulary` is called on page mount so the
  drawer is gated by the trigger, not by an in-modal spinner. Applying
  facets commits the draft search params and navigates with `page=1`.
- Tests for the drawer (open/close, focus, dirty-gating, apply payload),
  the RefineButton, the `parseFacets` round-trip, and an integration
  test covering the full facet flow on `SearchPage`.

## What's explicitly missing for now
* Refine button is ugly and not where it should be as per prototype
* Cancel button is not where it should be as per prototype 
* Reset does not trigger navigation (requirement on ticket) but simply resets the search draft.

### DIFF
--- a/src/components/filters/FilterDrawer.css
+++ b/src/components/filters/FilterDrawer.css
@@ -1,0 +1,115 @@
+.filter-drawer {
+  position: fixed;
+  inset: 0;
+  z-index: 100;
+}
+
+.filter-drawer__backdrop {
+  position: absolute;
+  inset: 0;
+  background: rgba(22, 27, 34, 0.4);
+}
+
+.filter-drawer__panel {
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  width: min(420px, 100vw);
+  background: var(--color-bg-surface);
+  border-left: 1px solid var(--color-border);
+  box-shadow: var(--shadow-lg);
+  display: flex;
+  flex-direction: column;
+}
+
+.filter-drawer__header {
+  padding: var(--spacing-md) var(--spacing-lg);
+  border-bottom: 1px solid var(--color-border);
+}
+
+.filter-drawer__title {
+  margin: 0;
+  font-size: var(--font-size-xl);
+  font-weight: 600;
+  color: var(--color-text-primary);
+}
+
+.filter-drawer__panel:focus {
+  outline: none;
+}
+
+.filter-drawer__body {
+  flex: 1;
+  overflow-y: auto;
+  padding: var(--spacing-md) var(--spacing-lg);
+  display: flex;
+  flex-direction: column;
+  gap: var(--spacing-md);
+}
+
+.filter-drawer__footer {
+  display: flex;
+  align-items: center;
+  gap: var(--spacing-sm);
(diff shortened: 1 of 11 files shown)


jaybea (approved):
I didn't find anything blocking the PR. 

One small race-conditioned fix might be worth taking on top of this branch because it's single new `useRef` plus a few lines inside an existing effect. Most of its value is future-proofing; the other observations are non-blocking polish so out of scope for this basic implementation.

**Optional, worth considering:**
User modifications to facets are lost when anything causes SerachPage to re-render while the drawer is open, like:
* Slow serach ressolves while the user has the drawer open refining filters.
* `userSearch.retry()` fires (the "Try again" button on the failed-search state)
* `useCorpusTotal` resolves on initial page load if the drawer is opened quickly (ok, that's really reaching...)
* Any future hook added to SearchPage that updates state asynchronously (facet counts - #73 and #74)

I repro'd it by sneakily tabbing out of the Drawer and hitting Space-bar to show the next page.

Real-world odds of seeing this exteremely low. 

**Future polish**
Focus is not trapped inside the modal, keyboard users can navigate to the underlying form with the drawer still open.
Keyboard users can't focus on a disable button, so "Loading filter..."/"Filters unavailable" tooltip is unreachable for keyboard users.

Just FYI: I was able to test the URL length, and above 150 single-concept facet fragment (!!!!), server started returning HTTP 431.

jack-o-the-green:
Great callouts, thanks @jaybea! I will have a quick look at that race-condition. 

Agree on the future polish - I'm going to do a ticket for (a) improvements to the UI and (b) accessibility fixes so will be sure to bundle that in. 

> Just FYI: I was able to test the URL length, and above 150 single-concept facet fragment (!!!!), server started returning HTTP 431.

This is a good find! I think we're safe with that limitation for now, I have on my list to raise a ticket for doing URI compaction using the vocabulary context (and potentially unpacking it on destiny-repository side using the same context) which will help with the length of what we're sending.

PR #85, opened 2026-05-25
Minimal UI polish for usability

jack-o-the-green:
Polish pass on the Filtering UI: relocate the Refine entry point, integrate the drawer's section cards with the panel, tidy up the meta-bar controls, and trim a few vocabulary schemes from the visible filter list.

## UI / cosmetic
- Move the Refine button from the search bar's year-range row into the results meta bar next to Sort, with a new `FilterIcon`.
- Move the drawer's Cancel button into the header (top-right) and rename the title/footer copy ("Refine the evidence", "Reset all", "Show results");
- Drop the per-card border, background, radius and shadow on FilterCards so they read as in-place sections separated by single dividers.
- Replace the Unicode `▸`/`▾` expansion glyphs with `ChevronRightIcon` / `ChevronDownIcon` SVGs that size to `1em` and inherit `currentColor`.
- Trim a trailing ` Scheme` from FilterCard titles so e.g. "Document Type Scheme" reads as "Document Type".
- Switch the Refine/Sort/Export controls to `--color-bg-surface` so they pop as clickable against the inset meta bar, and bump padding-y on all three to a common 7px so their heights align.
- Prefix every Sort option with `Sort: ` so the closed dropdown self-describes next to Refine.

## Behaviour
- Keep the FilterCard summary (e.g. "1 selected") visible when the card is expanded, separated from the title by a `·`.
- Keep the results meta bar visible when a Refine trigger is offered, so the button stays reachable in browse mode before the first result loads.
- Hide `EstimateSourceScheme` and `esea:ImplementationDescriptionScheme` from the drawer alongside the existing `EffectSizeMetricScheme` exclusion.
- Shrink the Refine count badge so it sits within the text line box — the button no longer changes height when filters are applied.
- Refine badge now counts individual selected concepts across the whole drawer (sum of each scheme's `selectedCount`) instead of the number of schemes with any selection — selecting 5 outcomes plus 1 document type now reads `6`, not `2`.

### DIFF
--- a/src/components/filters/FilterCard.tsx
+++ b/src/components/filters/FilterCard.tsx
@@ -1,5 +1,6 @@
 import type { ComponentChildren } from "preact";
 import { useId, useState } from "preact/hooks";
+import { ChevronDownIcon, ChevronRightIcon } from "@/components/icons";
 import "./FilterCard.css";
 
 interface FilterCardProps {
@@ -27,7 +28,7 @@ export function FilterCard({ title, summary, children }: FilterCardProps) {
           <span class="filter-card__summary">{summary}</span>
         )}
         <span class="filter-card__arrow" aria-hidden="true">
-          {expanded ? "▾" : "▸"}
+          {expanded ? <ChevronDownIcon /> : <ChevronRightIcon />}
         </span>
       </button>
       <div id={panelId} class="filter-card__panel" hidden={!expanded}>
--- a/src/components/filters/FilterCard.tsx
+++ b/src/components/filters/FilterCard.tsx
@@ -11,7 +11,7 @@ interface FilterCardProps {
 export function FilterCard({ title, summary, children }: FilterCardProps) {
   const [expanded, setExpanded] = useState(false);
   const panelId = useId();
-  const showSummary = !expanded && !!summary;
+  const showSummary = !!summary;
 
   return (
     <div class="filter-card">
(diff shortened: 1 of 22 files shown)


jack-o-the-green:
Fourish more commits coming after a chat with Anita, feel free to hold off reviewing until they land.

Update: Ready to go!

Adam-Hammo:
My definition pop-up is not escaping the drawer, so gets cut off:

image

src/components/filters/FilterDrawer.tsx
```diff
@@ -141,22 +149,31 @@ export function FilterDrawer({
...
         </header>
 
         <div class="filter-drawer__body">
           {schemes
             .filter(
               (scheme) =>
-                scheme.uri !== "evrepo:EffectSizeMetricScheme",
+                scheme.uri !== "evrepo:EffectSizeMetricScheme" &&
+                scheme.uri !== "evrepo:EstimateSourceScheme" &&
+                scheme.uri !== "esea:ImplementationDescriptionScheme",
```
Adam-Hammo: Nit: can we pull this out into a list somewhere less-deep?

src/components/filters/FilterCard.tsx
```diff
@@ -11,7 +12,7 @@ interface FilterCardProps {
 export function FilterCard({ title, summary, children }: FilterCardProps) {
   const [expanded, setExpanded] = useState(false);
   const panelId = useId();
-  const showSummary = !expanded && !!summary;
+  const showSummary = !!summary;
```
Adam-Hammo, line 15: javascript moment

src/components/filters/FilterDrawer.tsx
jack-o-the-green: Yeah we can pop it in to the community

Adam-Hammo (approved):

Issue #87, opened 2026-05-26
Add aggregations to CountryFilter


jack-o-the-green:
## Context 

Once [destiny-repository#710](https://github.com/destiny-evidence/destiny-repository/issues/710) is available, we can add aggregations to our CountryFilter implemented in #80 and also only include countries in our list that will produce results. 

<img width="658" height="743" alt="Image" src="https://github.com/user-attachments/assets/3a21bedc-0ffd-4fe5-a147-a3e8d5d9253f" />

Prototype https://claude.ai/design/p/4df77d9c-8dec-45f9-9a19-65cbce3b7cab?file=Fast+follow+prototype.html

## Technical Scope 

* Facet counts appear next to countries in the CountryFilter 
* Countries are only included in the filter if they are present in the aggregation, with a message that there are no references for a country if the user searches for something that isn't in the list. 
    * If this becomes a time sink, skip it.

PR #88, opened 2026-05-26
Country Filter

### LINKED ISSUE #80: Add country filter to FilterDrawer
## Context 

After the implementation of destiny-repository#696 we can add a country filter to FilterDrawer.

Fast follow prototype for country filter https://claude.ai/design/p/4df77d9c-8dec-45f9-9a19-65cbce3b7cab?file=Fast+follow+prototype.html

Image

Note that until destiny-evidence/destiny-repository#710 is done we will not be able to provide country aggregations, this is fine. 

## Technical Scope 

* Add a new CountryFilter that can be wrapped in a FilterCard and added to the FilterDrawer component. 
* This new filter will need to generate its lucene query syntax to be added into the query field
    * This should follow the same contract as the ConceptSchemeFilter which allows the FilterDrawer to construct a full query by having each filter generate its own query fragment. 
* Filter state should be repopulated from url on open (same as for concept search)

jack-o-the-green:
Closes #80.

## Summary

Adds a Country card to the FilterDrawer that lets users narrow searches by
ISO 3166-1 country. 

* Because country list is long, allow users to search for country. 
* Selected country narrows facet counts on concept scheme filters. 

Screenshot 2026-05-27 at 12 18 52

## Out of scope (intentional)

- **Per-country counts** pending destiny-repository #710


scripts/generate_countries.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def canonical_iso2(raw: str) -> str:
+    match = ISO2_RE.search(raw)
+    if match is None:
+        raise ValueError(f"no ISO 3166-1 alpha-2 code in {raw!r}")
+    return match.group(1)
+
+
+def main() -> None:
```
jack-o-the-green, line 34: Interested in thoughts on this, the Intl package used elsewhere has more display names/codes than country converter (and some disagreements), but country converter was used as our source of truth everywhere else. It felt appropriate to only include codes/countries in the map that our users could feasibly search for.

Adam-Hammo: When we get country aggregated counts could we actually populate this list from that?

jack-o-the-green: Yeah I think we should! It would reduce the length of the list too. We'll have communicate somehow that if a country is missing there are no results for it.

jack-o-the-green: Added it to the scope of #87

src/components/filters/countryFilterState.ts
```diff
@@ -0,0 +1,105 @@
...
+  return brand(new Set());
+}
+
+export function countryStateFromCodes(
+  codes: Iterable<string>,
+): CountryFilterState {
+  return brand(new Set(codes));
+}
+
+export function selectedCodes(state: CountryFilterState): readonly string[] {
```
danielbreves, line 22: A few of these duplicate conceptSchemeFilterState.ts helpers. Could we extract them to a utils/helper?

jack-o-the-green: My gut says it's too early to do this - we're going to make significant changes to the country implementation off the back of aggregations, and I prefer to abstract at a third instance of repetition. 

Happy to be disagreed with though!

danielbreves (approved):

PR #89, opened 2026-05-26
Naive facet counts

### LINKED ISSUE #73: Add facet counts to filter drawer

Adam-Hammo:
Closes #73 

Adds the naive facet counts from https://github.com/destiny-evidence/destiny-repository/pull/705 to the UI drawer.

Fast follow of https://github.com/destiny-evidence/destiny-repository/issues/703 will vastly improve the UX, but this provides a solid base for initial exploration.

### DIFF
--- a/src/components/filters/ConceptSchemeFilter.tsx
+++ b/src/components/filters/ConceptSchemeFilter.tsx
@@ -64,7 +64,7 @@ function ConceptItem({
   // rollup of descendants), which can confuse readers since clicking the
   // parent cascades into its subtree. Surface that semantic via a tooltip.
   const countTooltip = hasChildren
-    ? "Investigations tagged with this concept. Select to also include narrower concepts."
+    ? "Count shows investigations tagged with this concept. Select to also include narrower concepts."
     : undefined;
   const countNode = count !== undefined && (
     <span
--- a/src/components/filters/ConceptSchemeFilter.tsx
+++ b/src/components/filters/ConceptSchemeFilter.tsx
@@ -30,15 +30,12 @@ interface ConceptItemProps {
   onChange: (next: ConceptSchemeFilterState) => void;
 }
 
-// Compact-locale notation (1234 → "1.2K", 1_500_000 → "1.5M"). Honours the
-// browser locale via Intl rather than rolling our own k/M suffixes.
-const compactFormatter = new Intl.NumberFormat(undefined, {
-  notation: "compact",
-  maximumFractionDigits: 1,
-});
+// Locale-aware grouping (1234 → "1,234" in en-US, "1.234" in de-DE). Plain
+// Intl with no notation override; the browser picks the right separator.
+const countFormatter = new Intl.NumberFormat();
 
 function formatCount(n: number): string {
-  return compactFormatter.format(n);
+  return countFormatter.format(n);
 }
 
 function ConceptItem({
@@ -72,7 +69,7 @@ function ConceptItem({
   const countNode = count !== undefined && (
     <span
       class={countClass}
-      aria-label={`${count} references`}
+      aria-label={`${formatCount(count)} investigations`}
       tabIndex={hasChildren ? 0 : undefined}
     >
       {formatCount(count)}
--- a/src/components/filters/ConceptSchemeFilter.tsx
+++ b/src/components/filters/ConceptSchemeFilter.tsx
@@ -67,7 +67,7 @@ function ConceptItem({
   // rollup of descendants), which can confuse readers since clicking the
   // parent cascades into its subtree. Surface that semantic via a tooltip.
   const countTooltip = hasChildren
-    ? "References tagged with this concept. Select to also include narrower concepts."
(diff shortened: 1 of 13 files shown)


src/components/filters/ConceptSchemeFilter.tsx
```diff
@@ -37,6 +56,25 @@ function ConceptItem({ concept, state, index, onChange }: ConceptItemProps) {
...
   const hasChildren = !!concept.narrower && concept.narrower.length > 0;
+  const count = counts?.get(concept.uri);
+  const countClass = `concept-scheme-filter__count${
+    countsLoading ? " is-updating" : ""
+  }${hasChildren ? " concept-scheme-filter__count--parent" : ""}`;
+  // Parent rows expose only the count for the parent URI itself (not a
+  // rollup of descendants), which can confuse readers since clicking the
+  // parent cascades into its subtree. Surface that semantic via a tooltip.
+  const countTooltip = hasChildren
+    ? "Investigations tagged with this concept. Select to also include narrower concepts."
```
danielbreves: I thought this seemed a bit unclear without the context above. How about this?
```suggestion
    ? "Count shows investigations tagged with this concept. Select to also include narrower concepts."
```

src/components/filters/FilterDrawer.tsx
```diff
@@ -70,6 +76,20 @@ export function FilterDrawer({
...
+  const draftFacets = useMemo(
+    () => draftToFacets(draft, schemes),
+    [draft, schemes],
+  );
+  const facetParams: SearchParams = {
+    ...params,
+    searchFacets: open ? draftFacets : appliedFacets,
+  };
+  const { counts: facetCounts, loading: facetCountsLoading } =
+    useSearchFacets(facetParams);
```
danielbreves: Should we be checking and surfacing errors here?

src/components/filters/FilterDrawer.tsx
danielbreves: I wonder if we should have a `FilterDrawer` wrapper that prevents the hooks from running when it's closed?

```
export function FilterDrawer(props: FilterDrawerProps) {
  if (!props.open) return null;
  const { open: _open, ...rest } = props;
  return <FilterDrawerPanel {...rest} />;
}
```

jaybea (approved):
Wow, this works great! So awesome to be able see the counts to guide selections, even naive counts.

I put it through it's paces and couldn't find any issues with the naive facet counts, it's very solid. (there's a couple of pre-existing tiny issues, like you can tab out of the drawer, or if you enter an invalid year (like 2020-2010), bring up Refine, select a facet, then click Show Results, nothing happens.)

Tiny observation, under Study Design I was fooled by the lack of counts till I figured out blank means 0. 

Of course I clicked all the Study Design's facets and expected it to still include all the results, but got 0. 

Controversial, but maybe show a 0 on the facets that have no results, you might have an ugly wall of 0's.  Or just leave it, we'll be reworking this display code in a fast-follow I believe.

src/components/filters/FilterDrawer.tsx
Adam-Hammo: image
Have added this indicator if a request fails.

jack-o-the-green:
This looks great! I have one thought on usage. 

If I type a search term into the search bar, then open the filter drawer _before_ submitting, I get facet counts for all references, but then if I show the results, I get my search term + applied facets which can be significantly smaller. 

I am wondering if we want to include the search draft from outside the filterdrawer when sending our request for facet counts, as this would reflect the results the user will get?

Adam-Hammo:
> If I type a search term into the search bar, then open the filter drawer before submitting, I get facet counts for all references, but then if I show the results, I get my search term + applied facets which can be significantly smaller.
> 
> I am wondering if we want to include the search draft from outside the filterdrawer when sending our request for facet counts, as this would reflect the results the user will get?

My instinct is that yes, we should apply the search string to the facet counts, because clicking "Show Results" in the drawer will apply the string as well. Good catch!

Paging @acdonald who I trust far more than myself

Adam-Hammo:
Decision made to apply the draft search, will be done in a follow-up PR.

PR #95, opened 2026-05-27
Move year range filtering into filter drawer

### LINKED ISSUE #74: Move year range filter from SearchBar to FilterDrawer

jack-o-the-green:
## Summary 

Closes #74 

Moves the year range filter from the SearchBar into the FilterDrawer to consolidate our filtering into a single component. SearchBar is now only concerned with text based query from the user. 

## What's changed?

* Added new YearRangeFilter component inside a FilterCard to our FilterDrawer, this now handles filtering by a year range
    * Have preserved previous behavior of allowing just a start or just an end year.  
* YearRange and Country filters now start out expanded, as per our prototype


Adam-Hammo:
image

Just a visual nit on the size of the inputs - can we size the inputs closer to the size of 4 numbers?

src/components/filters/YearRangeFilter.tsx
```diff
@@ -0,0 +1,53 @@
...
+          autoComplete="off"
+          placeholder="YYYY"
+          aria-label="Start year"
+          aria-invalid={error !== null}
+          class="year-range-filter__year"
+          value={state.start}
+          onInput={(e) =>
+            onChange({ ...state, start: (e.target as HTMLInputElement).value })
+          }
+        />
```
Adam-Hammo, line 43: These inputs trigger `useSearchFacets` every keystroke as the draft search changes. Can we adjust to only update `facetParams` if the year is 4 digits long?

jack-o-the-green:
> Just a visual nit on the size of the inputs - can we size the inputs closer to the size of 4 numbers?

I'm not opposed but it will be a specific deviation from the prototype 
Screenshot 2026-05-27 at 15 00 23

src/components/filters/YearRangeFilter.tsx
jack-o-the-green: Nice catch, definitely!

Adam-Hammo:
Fair let's push it to a later polish if anything then

src/components/filters/YearRangeFilter.tsx
jack-o-the-green: Fixed, and adjusted the displayed error behavior to be more user friendly :)

Adam-Hammo (approved):
Looks good!

PR #98, opened 2026-05-28
Handle sibling counts in facet drawer

### LINKED ISSUE #87: Add aggregations to CountryFilter

### LINKED ISSUE #96: Add sibling count handling
After https://github.com/destiny-evidence/destiny-repository/issues/703, we can now correctly calculate sibling relationship counts for concepts.

UI plumbing changes we need to make:
- pass vocab uri to facet counts
- use structured concept query parameter for searching and faceting

UI changes we can make off the back of that:
- unsuppress sibling concept counts
- ~add total result count to the "apply filter" button~
- disable 0-count facet filters

Adam-Hammo:
Closes #96 

- Unredacts sibling counts with correct values
- Disables zero-count concepts
- Uses structured concept parameter

There was significant overlap, including transient code, so I've implemented #87 as well

- Adds facet counts to countries
- Populates countries from aggregation
  - Uses Intl to go from two-letter code to display name, but maintains source of truth at the repository 
- Uses structured country parameter


src/components/filters/ConceptSchemeFilter.tsx
```diff
@@ -72,11 +85,12 @@ function ConceptItem({
...
   return (
     <li class="concept-scheme-filter__item">
-      <label class="concept-scheme-filter__row">
+      <label class={rowClass}>
         <input
           class="concept-scheme-filter__checkbox"
           type="checkbox"
-          checked={isSelected(state, concept.uri)}
+          checked={selected}
+          disabled={isEmpty}
```
danielbreves, line 88: Would it be worth showing a tooltip to indicate why this is disabled?

tests/components/filters/FilterDrawer.test.tsx
```diff
@@ -63,7 +75,8 @@ describe("FilterDrawer", () => {
       <FilterDrawer
```
danielbreves: Could we write an overridable test helper for rendering `FilterDrawer`?

src/components/filters/conceptSchemeFilterState.ts
```diff
@@ -117,24 +117,31 @@ function walkConcepts(concepts: Concept[]): Concept[] {
...
-// Builds one SearchParams.searchFacets[] entry: an unwrapped, OR-joined
-// sequence of linked_data_concepts:"..." clauses for every selected
-// concept, with full concept URIs embedded verbatim.
-export function toSearchFacet(
+// Buckets selections into sibling-set groups for the backend's `concept=`
+// param: one group per common broader URI, with top concepts keyed by the
+// scheme URI so they share a synthetic root. Auto-rollup can leave a parent
+// + every descendant selected; they land in separate groups (disjoint sibling
+// sets), which is what the backend validates.
+export function toConceptFilterGroups(
```
danielbreves, line 97: It seems the auto-rollup can lead to some odd behaviour. If a parent has no documents, but all its children do, clicking all its children will auto select the parent with no documents, which leads to 0 results, even though the user only selected concepts with counts.  

Image

Similar to this, a parent can report more results than what you get when selecting it, since selecting it will AND it with all its children.

Image
Image

I wonder if the backend could expand the children when a parent is selected. But perhaps we could just decouple the parent children selection in both directions for now?

src/components/filters/CountryFilter.tsx
```diff
@@ -34,21 +76,43 @@ export function CountryFilter({ state, onChange }: CountryFilterProps) {
...
+                )}
+              </label>
+            </li>
+          );
+        })}
         {visible.length === 0 && (
-          <li class="country-filter__empty">No countries match.</li>
+          <li class="country-filter__empty">
+            {counts == null
+              ? "Loading countries…"
```
danielbreves, line 111: Should this appear when `countsLoading` rather than `counts == null`

src/components/filters/conceptSchemeFilterState.ts
Adam-Hammo: This is a really good point.

For now I'd agree that we decouple the selections. At the root, parents AND-ing with children in the query is an existing bug. This will align counts with that behaviour, for better or worse. Decoupling will at least allow the user to see the records of a parent or all of its children.

We probably need to circle back to expected behaviour with parents and hierarchies across the board - tagging https://github.com/destiny-evidence/destiny-repository/issues/655 and https://github.com/destiny-evidence/destiny-repository/issues/712.

src/components/filters/ConceptSchemeFilter.tsx
Adam-Hammo: Where would you see this sitting on concepts that already have definition tooltips?

danielbreves: Good question, in the same tooltip perhaps? 🤷 It's just a UX nit, so happy to leave this out.

src/components/filters/conceptSchemeFilterState.ts
```diff
@@ -70,41 +70,17 @@ export function buildConceptIndex(scheme: ConceptScheme): ConceptIndex {
...
-// then reconciles ancestors upward via `index.broader`: at each level the
-// parent is selected iff every one of its `narrower` is selected. Recomputing
-// from sibling state handles both directions in one pass — selecting the last
-// missing sibling rolls a parent in, deselecting any sibling rolls it back out.
-// Direction of the initial subtree change is set by the clicked concept's
-// current state: if selected, the whole subtree is cleared; otherwise added.
-// This loses any independent child selections on deselect — a deliberate
-// tradeoff for predictable subtree semantics.
+// Toggles a single concept URI. Subtree cascade and ancestor auto-rollup
+// previously lived here, but interacted badly with the backend's literal-URI
```
danielbreves: Nit: it seems a bit odd to refer to code that isn't here anymore... 🪦

danielbreves (approved):
LGTM!

src/components/filters/conceptSchemeFilterState.ts
Adam-Hammo: Fair, it's a pet hate of mine too! I missed it

PR #104, opened 2026-05-29
Add fonts

### LINKED ISSUE #90: Update Fonts

Adam-Hammo:
Adds and maps Untitled Sans regular and bold, and Signifier Medium.

Net effect on the UI is that it looks thinner.

See https://covidence.slack.com/archives/C0810ARPCRG/p1780015213598879?thread_ts=1779769211.067839&cid=C0810ARPCRG


jaybea (approved):
Approve after one should-fix. Every label in the Refine drawer flips from Medium to Bold as an implicit consequence of shipping the Bold cut. Demoed to the team and the consensus was to keep medium, so pin the two title rules to 500. Everything else checks out, CI is green.

PR #115, opened 2026-06-03
Organise components by domain (investigation, finding, shared)

danielbreves:
Groups loose components in `src/components/` into domain folders so the directory is navigable as it grows. Tests mirror the new structure.

- `investigation/` — InvestigationCard, AbstractSection, SharedContextBlock, FindingsSection
- `finding/` — FindingCard, ComparisonRow, EffectEstimateCard, ArmDataTable, CIVisualization, InterventionDetails, ContextDetails, SampleDetails, OutcomeDetails, SourceEvidenceToggle
- `common/` — Tooltip, TagGroup, LabeledField, Pagination, icons

Existing `filters/`, `search/`, `layout/`, `feedback/` are unchanged. All moves via `git mv` so history follows; typecheck clean; all 670 tests pass.

This is opinionated and open to suggestions — happy to revisit:

- **investigation/finding boundary.** `InterventionDetails` / `ContextDetails` are used both per-finding and by `SharedContextBlock` (investigation-level). I put them in `finding/` since they describe the parts of a finding; the alternative is calling them investigation-level primitives.
- **`Pagination` in `common/`.** Only used by search results today; could equally live in `search/`.
- **Naming.** `finding/` (singular) to match `investigation/`; could also be `findings/`.


src/components/common/icons.tsx
Adam-Hammo, line 1: If we're dropping opinions, I tend towards `common` over `shared` personally 🙈

Adam-Hammo (approved):

PR #122, opened 2026-06-03
Plumb in cross-facet api

### LINKED ISSUE #117: Wire up cross-facets endpoint (API client + data hook)
Scope: #93. API: destiny-evidence/destiny-repository#727 (merged).

Add the client + data-fetching layer for the new cross-facets endpoint so the map has data to render.

**Endpoint:** `GET /references/search/cross-facets/` — same filters as `/references/search/`, plus `row` and `column` (each one of `countries`, `country_wb_regions`, or a concept-scheme URI) and an optional `vocabulary` URI (required when an axis is a scheme URI). Returns `{ total, cells: [{ row, column, count }] }`, non-zero cells only.

**Needs**
- `api.crossFacets(...)` in the API client + response types mirroring the SDK (`CrossFacetCell`, `ReferenceCrossFacetResult`).
- A fetch hook in the style of the existing `useSearch*` hooks.
- Mapping from an "axis selection" to the `row`/`column`/`vocabulary` query params.

**Gotchas** — cells can sum to more than `total` (a reference can carry multiple values on an axis); use `total`, don't sum cells. Pure data layer, no UI dependency.

**Done when** — given a filter set + axis pair, the hook returns typed cells + total from the endpoint, verifiable in isolation (test or throwaway harness) without any map UI.

Adam-Hammo:
Closes #117 by adding the plumbing for the incoming `/cross-facets/` endpoint.

Cannot be merged until https://github.com/destiny-evidence/destiny-repository/pull/727.

### DIFF
--- a/src/hooks/useCrossFacets.ts
+++ b/src/hooks/useCrossFacets.ts
@@ -1,0 +1,89 @@
+import { useEffect, useState } from "preact/hooks";
+import { crossFacets } from "@/services/apiClient";
+import { useCommunity } from "@/community/CommunityContext";
+import { toTurtleUrl } from "@/services/vocabulary/vocabularyService";
+import {
+  axisPairToParams,
+  type CrossFacetAxisPair,
+} from "@/services/crossFacets";
+import type { SearchParams } from "@/services/searchParams";
+import type { ReferenceCrossFacetResult } from "@/types/models";
+
+function paramsKey(
+  params: SearchParams,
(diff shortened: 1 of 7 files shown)


jaybea (approved):
Solid, well-tested work that faithfully follows the established facets patterns. I verified end-to-end against the live backend over the ESEA corpus.

Wish we had HPV data, but the live verification exercised and validated all paths in this PR anyway.

PR #123, opened 2026-06-04
Evidence map

### LINKED ISSUE #118: Build the evidence map grid (Table + Bubble views)
Scope/wireframes: #93.

The core canvas: a cross-tabulation matrix of two taxonomy dimensions, rendered from the cross-facets cells.

**Needs**
- Matrix from the returned cells; row labels + column headers derived from the selected concept schemes, and sticky.
- **Bubble view:** count-sized bubbles (bigger = more records) + a legend/scale at the bottom of the canvas. 
  - Adam opinion: bubbles should scale by area not diameter.
- **Table view:** count as text.
- Bubble/Table toggle; total record count (e.g. "247 investigations"); cell hover tooltip ("Click to view matching investigations").
- **Empty / over-filtered state:** "0 records", inline warning banner with a "Reset all" CTA, grid still rendered with cells greyed/empty.
- Default axes: WHO region on y, thematic focus on x.

**Notes** — wireframes are indicative; reuse existing components/styles rather than matching fonts/looks. Build the grid to render from supplied data + axis config (the data hook with default axes, or a fixed filter set) so it isn't blocked on the configure panel — live filter/axis controls arrive separately.

**Done when** — both views render correctly from supplied data (bubbles + legend + total + tooltip; table counts), headers are sticky and labelled from the selected schemes, and the empty/over-filtered state displays — all driven by the default axes/stub inputs, without requiring the configure panel.

Adam-Hammo:
Closes #118 

Adds URL-driven evidence map (bubble & table view) to the visualise tab.

At a (very) high-level the diff contains:
- a small amount of code shuffling to reuse logic (eg scheme URI -> labels)
- css/tsx/ts to build the evidence map

Main design callout: all concepts in the scheme are present, regardless of whether they are parents, empty, etc
> _It's a little awkward how parent concepts display... they could look empty (a false "gap") when the children are populated, or vice-versa. They're also all sorted alphabetically, so don't live next to their children. I'm not sure that's something we can solve now, we need a holistic approach across counts/searches/maps here and it hurts my brain to think about too much. Right now they just show the count of references directly coded to them.)_

> One other small callout: countries, when presented as properties, won't show missing rows/columns - as we don't have the canonical list of countries on the UI, as they're not enumerated the vocabulary. I.e. only a minor problem for ESEA, not HPV. Given our conversation on Thursday it's unlikely that showing all of the missing countries is a common user pattern anyway.

Easiest way to test yourself: run locally against staging API and `git revert 3280d36` to toggle it on for ESEA data.


src/services/vocabulary/vocabularyService.ts
```diff
@@ -94,29 +105,39 @@ interface RawScheme {
  *
  * SKOS allows polyhierarchy; `broader` keeps only the first parent and the
  * scheme tree mirrors that single-parent view.
+ *
+ * All `@id`s are expanded to full IRIs via the document's own `@context` prefix
```
jack-o-the-green: Is `IRI` intentional over `URI`? I see it's used twice which makes me think I'm missing something.

jack-o-the-green (commented):
Overall looking really good! Primarily behavioral comments from me

(1) If you switch tabs to visualize, any filters you have applied to your search are wiped and not there when you come back to the search tab. Additionally, applied filters from the visualize tab are not maintained on tab switch on #128. Appreciate this is somewhat tricky, but I wouldn't be surprised if we have an initial user pattern of 

(a) filter some results on the search page
(b) click "visualize" with the expectation that they'll see a visualization of that search
(c) realize they don't see a visualization of that search (or worse, not realize the search isn't applied).
(d) go back to the search page because they want those results 
(e) realize their filtering is gone and get mad because they have to put it in again. 

it would be nice to be able to persist a query between the search and visualize tabs, or at least maintain the search on the search tab independent of what's happening on the visualize tab

(2) (Feel free to ignore) It would be nice to have the row and column highlight on hover so that you have a really clear indicator of what row/column you're over. Am thinking of instances of where the rows/columns could get quite large. Probably something for later. 

(3) I'm unsure of the value in providing both bubble and table views, they're basically the same. I'd be keen to only provide bubble view in interests of simplicity, it's by far the best visually. 

(4) More musing for the future, but I expect we'll have more than two-tier hierarchies in the destiny taxonomy when it's ready. I think the parent child relationships on the evidence map will get tricky at that point. I do think alphabetical sorting is the best way to go for now though.

jaybea:
+1 on Search -> Visualise not carrying over. I did exactly that: put in the standard "music", got 111 results, wanted to see that visually, and whoa, 3,000+ across all the countries.

I dug into the specs and the prototype and it never says it's carried over - if anything, reading the prototype left to right implies it isn't ("afghanistan" search, 15 refs, shows 247 on Visualise). So the map is a standalone, default-configured view: it ignores the Search q/filters and loads the whole corpus, which is why 3,000 and not 111 (and the per-country counts are whole-corpus facets, not the music subset).

That said, +1 that it's human nature to expect it to carry, and our UI - the tab sitting right next to Search - probably encourages the assumption. I think this could be done if we decide since cross-facets backend accepts q.

What IS specified is the reverse, and it's wired in #129: clicking a map cell deep-links INTO a pre-filtered Search (that cell's row x column become the filters), with a "Back to Visualise" return. So the linkage is intentional and one-directional - cell -> Search, not Search -> map.

src/services/vocabulary/vocabularyService.ts
Adam-Hammo: Nope, this is a slip

jaybea (approved):
Fantastic foundation. 

I've pruned any downstream(upstream)? Findings from #128 and #129 from this feedback.

Just a couple of things to ponder, roll into #129 if you think important enough. No blockers specific to this PR:

The corner <th class="evidence-map__corner" scope="col"> doesn't actually head a data column or a row. It carries the total and the axis legend. scope="col" makes assistive tools associate it with the column beneath it (the row-label column), so a screen reader announces the whole "8,500 investigations / Columns: … / Rows: …" block as that column's header. Dropping "scope" (or role="presentation") fixes it. Minor, but timely with the headers about to become clickable. (The ↔/↕ glyphs are already aria-hidden — nice.) 

**over-filtered banner test** asserts the fixture default, not the live `countNoun`.

Adam-Hammo:
Nice ones, applying to #129 and merging this

Issue #93, opened 2026-05-27
Evidence Maps UI Breakdown


acdonald:
### Evidence Maps
Wireframes: https://claude.ai/design/p/9d2290d4-6daf-4578-a7eb-bffc346adb72?file=Final+Designs.html

### Overview
- Build the Visualise experience, allowing users to generate an evidence map by selecting two taxonomy dimensions and optional filters. This is a core feature for helping users understand where evidence exists across the repository.
- A kickoff is needed to align on scope, confirm technical feasibility of the dimension/filter data model, and break this into epics.
- Wireframes shouldn't be taken as strict UI implementation. As such, ignore much of how the components/fonts look and reuse what we've got today

### Feature flag
- ability to turn this feature on/off by community i.e. HPV has evidence maps but Education doesn't

### Navigation tabs
- A persistent header tab bar displays "Search" and "Visualise" as the two primary navigation options
- The active tab is visually indicated with an underline
- Tabs are accessible from both the search view and the visualise view
- Clicking "Search" routes to the search experience; clicking "Visualise" routes to the evidence map

### Evidence maps - default/configured
- VISUALISE tab is active in the header
- By default, WHO region should be selected on the y-axis and thematic focus on the x-axis
- Bubble view: cells display sized bubbles representing the count of matching records; larger bubbles = more records
- Table view: cells display text representing the count of matching records
- A legend / scale is shown at the bottom of the canvas (e.g. 4, 12, 24, 0) when the bubble view is selected - ask Adam his opinion on this
- A total record count is shown (e.g. "247 investigations")
- Row labels (e.g. Sub-Sah. Africa, Middle East, Europe, Latin America, South Asia) and column headers (e.g. Education, Health, Governance, Environment, Economics) are rendered from the selected taxonomy concept schemes and should be sticky
- A hint is shown at top right: "Click a cell to view matching investigations"
- The right panel shows the configured axis dropdowns and an active Filters section. This should be the same Filters components as the search page
- "Reset all" and "Show results" buttons are shown at the bottom of the panel
- Bubble and Table toggle is active
- On cell hover the user should see a tooltip that says 'Click to view matching investigations' upon clicking it takes the user to the search page with those filters applied

### Evidence maps - no results (over-filtered)
- Record count shows "0 records"
- An inline warning banner reads: "No investigations match the current filters. Please update the filters and try again."
- A "Reset all" CTA is shown inline in the banner
[...]

PR #128, opened 2026-06-05
Evidence map filter

### LINKED ISSUE #119: Build the evidence map configure panel (axes + filters)
Scope/wireframes: #93. Reuses the search filters.

The right-hand panel that drives the map.

**Needs**
- Axis dropdowns for the row (y) and column (x) dimensions, populated with concept schemes and non-concept options eg (countries and regions).
- An active Filters section that **reuses the same filter components as the search page** (`FilterDrawer` / the existing `*FilterState` modules), not a reimplementation.
- "Reset all" and "Show results" actions at the bottom of the panel.

**Open question** — should the panel also include a free-text search box mirroring the search page, so the map can be narrowed by query as well as filters?

**Done when** — changing axes or filters in the panel updates the query and the map re-renders; "Reset all" clears selections; "Show results" applies them.

Adam-Hammo:
Closes #119 

- Refactors filter drawer to make it reusable
- Adds filters & axis selection to the right side of the visualisation page
- Some other CSS tweaks to the grid that became apparent when filtering
  - (Including aligning the no-result page to the scope in #93 ) 

image
*Pls ignore the incorrect bubble sizes in the screenshot, they have been fixed

image


src/pages/VisualisePage.tsx
```diff
@@ -159,63 +211,94 @@ function EvidenceMapView({
...
+                rows={model.rows}
+                columns={model.columns}
+                getCount={model.getCount}
+                maxCount={model.maxCount}
+                view={view}
+                countNoun={noun}
+                rowAxisLabel={rowAxis.title}
+                columnAxisLabel={columnAxis.title}
+                total={formatTotal(result.total)}
+                updating={loading}
```
jaybea, line 334: While `loading`, `result` still holds the previous axes's cells, but `model` is rebuilt against the new `rowAxiis/columnAxis`. So the grid renders the old result but these can't be resolved so they display as URI (admittedly, very briefly, user might not be able to even see this.) Filter-only changes don't have a problem, only axis changes. See the screenshot.

tests/components/visualise/MapConfigPanel.test.tsx
```diff
@@ -0,0 +1,169 @@
...
+      filters: {
+        conceptFilters: [[URI_JOURNAL]],
+        countryCodes: [],
+        startYear: undefined,
+        endYear: undefined,
+      },
+    });
+  });
+
+  test("Reset all restores the default axes and clears filter selections", () => {
```
jaybea, line 138: baseProps set `defaultAxes == appliedAxes == AXES`, so the reset test at line 138 checks whether `handleReset` reads `defaultAxes` or `appliedAxes`. It doesn't test the actual contract (Reset restores community defaults, not applied). Maybe add a case with `defaultAxes != appliedAxes`?

jaybea (approved):
This works great, I couldn't find any blockers. I love how the placeholders get replaced on a vocab change or one an initial slow-load of the vocab using the useMemo, it's SO elegant! 

A couple of findings that I think you'll want to fix (and maybe some not worth it.)

1. During an axis change, you **could** briefly see URL's down the left column, see attached screenshot. This actually happens on every axis change, you just need the refetch to be slow enough to see it.

Image


2. On non-Mac browsers (i.e. Windows), and it turns out on my Chrome with old-school settings where scroll-bars persist (and don't float and disappear), it makes for a little off by 7pixels or so on the un-docked layout below 960px. Pretty picky nit-pick.

3. Just a little consistency/weirdness with 2 "Reset all" on the screen, and they behave slightly different. I don't really know what the fix (if we decide to do anything) on this one is. If you "over-filter" to 0 results, we show the helpful message and a "Reset All" hyper-text. Of course the FilterPanel "Reset All" is still there. Click the hyperlink applies immediatly and ALSO resets the axes to defaults, while clicking "Reset all" in the panel footer only resets the draft and still needs "Show results".

Image

4. A slight gap in test coverage around "Reset all restores the default axes". It passes wither `handleReset` reads `defaultAxes` or `appliedAxes`. It doesn't seem to assert that Reset resetores the comunity defaults alone.

PR #129, opened 2026-06-05
Visualise<->Search navigation

### LINKED ISSUE #120: Deep-link from a map cell into pre-filtered Search
Scope/wireframes: #93.

Clicking a cell takes the user to Search, pre-filtered to that cell's row + column.

**Needs**
- Cell click navigates to Search with the cell's row + column applied as filters (reuse the existing search-URL / `searchParams` building).
- Active tab switches to Search on navigation.
- A "Back to Visualise" link at the top of the search results page to return to the map.

**Done when** — clicking a cell lands on Search pre-filtered to that cell's row + column with the Search tab active, and "Back to Visualise" returns to the map.

Adam-Hammo:
Closes #120 

Makes cells clickable to take you to the search page, with a "back to visualisation" button to take you... back!

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,8 +10,8 @@ import { URL_CHANGE_EVENT } from "./services/navigation";
 
 // preact-router intercepts internal <a href="/..."> clicks and updates the
 // URL via history.pushState without firing popstate or our URL_CHANGE_EVENT.
-// Bridging Router onChange into URL_CHANGE_EVENT keeps subscribers
-// (CommunityProvider, useUrlParams) in sync with router-driven navigation.
+// Bridging Router onChange into URL_CHANGE_EVENT keeps our URL subscribers
+// in sync with router-driven navigation.
 function emitUrlChange() {
   window.dispatchEvent(new Event(URL_CHANGE_EVENT));
 }
--- a/src/components/common/Tooltip.css
+++ b/src/components/common/Tooltip.css
@@ -32,7 +32,9 @@
   line-height: 1.4;
   width: max-content;
   max-width: 280px;
-  white-space: normal;
+  /* pre-line honours explicit "\n" in `text` (multi-line tooltips) while still
+     wrapping/collapsing other whitespace; single-line tooltips are unaffected. */
+  white-space: pre-line;
   text-align: left;
   z-index: 10;
   pointer-events: none;
--- a/src/components/common/Tooltip.css
+++ b/src/components/common/Tooltip.css
@@ -10,6 +10,16 @@
  *   so it fits inside the available space (see ConceptSchemeFilter.css).
  */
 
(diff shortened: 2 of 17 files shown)


jaybea (changes requested):
Very nice solution to manipulate and use the `history.state` rather than a query param.

A few things would be good to tighten this up.

1. **"Back to Visualise" disappears after the first Search interaction.** `navigate` clears `history.state` by default (`navigation.ts:10`, `state ?? null`), and none of SearchPage's own navigations preserve it: pagination (`SearchPage.tsx:222`), sort (`:227`), and submit (`:218`) all call `navigate(buildSearchUrl(...))` with no state. Maybe the last one, `submit` *should* clear it, whatever you think is appropriate.

2. **Stale cells stay clickable while the map is refetching.** `useCrossFacets` keeps the prevous result during a reload (`useCrossFacets.ts:46-78`), and the grid keeps `onCellClick` wired (`VisualisePage.tsx:294`), but `handleCellClick` holds onto current `params`/`axes`. Clicking during the refetch builds a search from the new filters and old cell data, especially if the axes changed mid-flight. Maybe guard the cells like the Export button is guarded.

3. **The cell button has no accessible action label.** A screen reader would just annouce "5, button". No `aria-label`.

4. **A country-axis click breaks the cell's "view matching investigations" count.**. I think the spec said clicking on a cell should "view matching investigations." That holds for a scheme-axis which will show the same count of result show on the cell. But in a certain siguation, with a country filter already active and then clicking on another country investigation, this will OR the countries, giving you a result that is *more* than the listed number on the cell. I don't know what the correct behaviour is, but it looks wrong to click on the 2nd cell which shows "2", and have 10 results shown.

### Nitpicks

`navigation.ts:16' comment is stale, there's a couple of other subscribers. same nitpick at `App.tsx:14`. Fine with me if they're not listed but described gnerically, or the list fixed up (or left alone, this is a nitpick!)

The back-link route guard is great, but it could be made tighter than "startsWith`, since that would allow: `/esea/visualise-co-opted-control`. (Oh come on, give me a break, that is really reaching...)

A small test gap, "a countries axis contributes a country code, de-duplicated" test assert only deduplication (the `FR` case). The additive path runs (`DE`) but nothing asserts (that I could find) it landed in `next.countryCodes`. `push` could be deleted on src/services/evidenceMap.ts:204 and ell 3 tests from evidenceMap.test.ts:304-357 would still pass.

Adam-Hammo:
Thanks @jaybea, addressed all except 4, which I've ticketed for further discussion #133, and 1, which we've decided to maintain pending user feedback for desired functionality (eg, if you filter, should "back" take you back to the original viz or the newly filtered one?)

jaybea:
Review feedback fixes look good! Ready to go from my viewpoint!

jaybea (approved):
Review feedback changes look good!

Issue #127, opened 2026-06-05
HPV Export to Excel: reference-level workbook for HPV community


danielbreves:
Part of the UI epic [#103](https://github.com/destiny-evidence/evidence-repo-ui/issues/103). Builds on [#106](https://github.com/destiny-evidence/evidence-repo-ui/issues/106).

Clicking "Export to Excel" on `/hpv` search results produces an `.xlsx` file with:

- One row per reference (no Findings or Outcomes sheets)
- Columns for bibliographic data (title, authors, year, journal, DOI, abstract)
- Columns for the applied concepts grouped by scheme

Filename follows the existing pattern: `destiny-evidence-hpv-YYYYMMDD.xlsx`.

What's different from Education's export:

- **No Investigation / Finding / Outcome sheet hierarchy** — HPV references don't have those.
- **Concept columns are derived from HPV's scheme list**, not Education's hard-coded scheme columns.

## What stays the same

- Education (`/esea`) export behaves exactly as today.
- Export request flow (request → poll → fetch → build) unchanged.
- Filename pattern, button location, visual treatment.

## Verify

- Trigger Export to Excel on `/hpv` with seeded data; result `.xlsx` has the expected sheet structure, one row per reference, concept columns populated.
- `/esea` export unchanged; existing tests pass.

## Pointers

- Export pipeline lives in [`src/services/export/`](src/services/export/). The current build is heavily Education-shaped — `SHEET_HEADERS` in `buildRows.ts` hard-codes Education-specific columns, and `generateWorkbook` builds Investigation/Finding/Outcome sheets via `buildInvestigationRow`, `buildFindingRows`, `buildOutcomeRows`.
- A per-community workbook builder dispatch is a natural factoring — keeps the Education flow untouched, adds an HPV variant alongside it.
- The backend export endpoint should be community-agnostic (it returns the same `Reference` + `Enhancement` shape regardless of community); only the client-side workbook-building changes. Worth a quick verification while implementing — if the backend turns out to make Education-specific assumptions, that becomes a small destiny-repository follow-up.

## Depends on

- [#106](https://github.com/destiny-evidence/evidence-repo-ui/issues/106) for HPV community routing.
- End-to-end verification depends on [destiny-repository#731](https://github.com/destiny-evidence/destiny-repository/issues/731) landing seeded HPV data.

Issue #106, opened 2026-06-01
HPV community: routing, config, and community-aware copy


danielbreves:
Part of https://github.com/destiny-evidence/evidence-repo-ui/issues/103.

## Summary

Add HPV Vaccine Delivery as a sibling community to the existing Education (`esea`) community, and parameterise the hard-coded Education-specific copy so `/hpv` and `/esea` can coexist without leaking strings between them. This is the foundation for the rest of #103 — nothing else can be tested under `/hpv` without it.

### Community wiring

- Add an `hpv` entry to `COMMUNITIES` in `src/services/communities.ts` with `slug: "hpv"`, `name: "HPV Vaccine Delivery"`, vocabulary/context URLs, `defaultAnnotations` (TBD with backend), and an empty `filterExcludedSchemes` for now.
- Add `VITE_HPV_VOCABULARY_URL` and `VITE_HPV_CONTEXT_URL` to `src/config.ts`, `.env.example`, and `tests/setup.ts`.
- Stand up an HPV vocabulary JSON-LD source — convert `taxonomy-builder/vocab/hpv/hpv-vaccine-delivery.ttl` (36 SKOS concept schemes) to JSON-LD. Commit a static file under `public/vocab/` or point at a hosted URL.
- Confirm `CommunityContext` resolves `/hpv` cleanly; Education flow at `/esea` is unaffected.

### Community-aware copy

Today many Education-specific strings are baked into component JSX. Centralise community-display strings on the `Community` type (`src/types/models.ts`) or a sibling `CommunityCopy` map — at minimum: hero subtitle template, search placeholder, results-count noun (\"references\" vs \"investigations\"), export-file slug, drawer title.

HPV values per design:

- Header: `Evidence Repository / HPV Vaccine Delivery`
- Hero: \"Search the evidence\" / \"{N}+ references across HPV vaccine delivery research\"

## Verification

- Visit `/hpv` in dev — header crumb reads \"Evidence Repository / HPV Vaccine Delivery\", every visible string matches the HPV design, and the HPV vocabulary loads (check Network tab for the JSON-LD request).
- Visit `/esea` — every string and behaviour is unchanged.
- Tests: `CommunityContext` resolution for `/hpv`, copy fallbacks per community.

Adam-Hammo:
While we're here, I'll feature flag the evidence maps per-community

Issue #103, opened 2026-04-28
Adapt Evidence Repo UI for HPV


mootpointer:
## Summary

Adapt the Evidence Repository UI so it works cleanly for the HPV delivery, replacing education-specific assumptions with HPV-specific taxonomy, labels, metadata, and browsing behavior.

## Goal

Enable users to search, browse, and understand HPV evidence in the repo using the new HPV taxonomy and associated metadata, without requiring education-specific structures or terminology.

## Context

-   The current UI is built around education data assumptions and needs to be adapted for HPV.
    
-   The HPV delivery depends on the UI working against HPV data and taxonomy, alongside maps and AI summaries (not in scope for this Epic).
        
-   Data extraction is expected to map directly to the new taxonomy, so the UI should be able to work from that taxonomy/code set cleanly.
    
-  The HPV taxonomy/codesets are mostly flat for now, as opposed to deeply structured. 

- In the near term, the UI likely needs to support less structured tags that are more floating in space than the education setup.

    

## Problem statement

The current repository UI reflects education-oriented data structures, labels, and assumptions. For HPV, users need an interface that can present HPV-specific concepts and filters in a way that feels coherent even if the underlying taxonomy is currently less structured than the education one.

## Scope

-   Update UI labels, terminology, and framing from education to HPV
    
-   Support HPV taxonomy concepts in search, filtering, and record display (see [prototype](https://claude.ai/design/p/edf38b42-9f20-44d9-aabb-582a016c0f49?file=HPV+Evidence+Repository.html)).    

## Out of scope

-   Building the evidence map itself
    
-   AI summaries implementation
    
-   Full taxonomy-builder integration for mapping    
    
## Open questions

-   Which existing education-specific components can be reused as-is vs. need refactoring?

jack-o-the-green:
If we want to support exports for HPV we will need a ticket for this, currently very hard-coded to Jacobs/ESEA

danielbreves:
cc @mootpointer @acdonald do we need exports for HPV?

acdonald:
yes we will want exports - ok if it's a fast follow after the learning convening

jack-o-the-green:
> yes we will want exports - ok if it's a fast follow after the learning convening

Okay! We should disable exports for the community/hide the button until it's available then. Otherwise they'll get something super weird if they click it.

PR #130, opened 2026-06-07
HPV community: result-card tags, evidence-map defaults, and per-community feature gates

### LINKED ISSUE #125: HPV result cards: render applied-concept tags, drop findings/estimates
Part of the UI epic [#103](https://github.com/destiny-evidence/evidence-repo-ui/issues/103). Builds on [#106](https://github.com/destiny-evidence/evidence-repo-ui/issues/106).

## User story

Viewing search results on `/hpv`, each result card shows:

- title, year, authors, two-line abstract preview, DOI
- the concepts coded against the reference, rendered as tags — capped at 8 with a "+N more" overflow

The results toolbar on `/hpv` does **not** show the Export to Excel button until the HPV export work in [#127](https://github.com/destiny-evidence/evidence-repo-ui/issues/127) lands — clicking it today would run the Education-shaped pipeline against HPV data and produce a malformed workbook.

The Refine drawer on `/hpv` also hides the polished Country filter (the typeahead picker backed by the dedicated `countries` search facet). HPV has no data in that facet yet — it codes country as an ISO suffix on a concept URI rather than the object property ESEA uses — so the picker renders an empty "No countries match the current filters." Hiding it removes that dead card; the flag is intended to flip back on for HPV as a fast-follow once the country projection populates the facet.

What's different from Education's result cards:

- **No "findings"/"estimates" labels** — HPV references don't have findings or effect estimates.
- **Tags come from the reference's applied-concept data** rather than being aggregated from scheme-specific Education fields.
- **Export to Excel button hidden** until [#127](https://github.com/destiny-evidence/evidence-repo-ui/issues/127) lands.
- **Polished Country filter hidden in the Refine drawer** — its `countries` facet is empty for HPV, so the picker would render with no countries.

## What stays the same

- Education (`/esea`) result cards and Export to Excel button behave exactly as today.
- Education's Refine drawer, including the polished Country filter with its facet counts, is unchanged.
- Tag rendering, overflow, and visual treatment match the existing component.

## Verify

- `/hpv` search returns cards that look like the four card mockups in the HPV design.
- The Export to Excel button is not visible on `/hpv` (no broken workbook reachable).
- `/hpv` Refine drawer shows no polished Country filter card; `/esea` still shows it populated with country facet counts.
- `/esea` cards and Export to Excel button unchanged; existing tests pass.

## Pointers

- Result card markup lives in [`src/components/search/ResultRow.tsx`](src/components/search/ResultRow.tsx).
- Tag aggregation flows through [`src/services/investigationParser.ts`](src/services/investigationParser.ts) — extending it to recognise HPV's applied-concept slot is likely the smallest change.
- The Export to Excel button lives in the results toolbar on [`src/pages/SearchPage.tsx`](src/pages/SearchPage.tsx). A community-config boolean (e.g. `community.features.export`) defaulting to `true` for ESEA and `false` for HPV gates the render — same shape of community flag as the findings/estimates badge guard in this issue.
- The polished Country filter is gated by a new `community.features.countryFacetFilter` boolean (ESEA `true` / HPV `false`), threaded into [`FilterDrawer`](src/components/filters/FilterDrawer.tsx) as `showCountryFacetFilter` — same shape as the export flag above. It **cannot** be gated via `filterExcludedSchemes`, which only filters the vocabulary concept-scheme cards built in `SearchPage.tsx`, not the hardcoded Country `FilterCard`.
- HPV linked-data shape and backend availability tracked in [destiny-repository#731](https://github.com/destiny-evidence/destiny-repository/issues/731).

## Depends on

- End-to-end verification depends on [destiny-repository#731](https://github.com/destiny-evidence/destiny-repository/issues/731) landing seeded HPV data + the projection fix.

jaybea:
Closes #125. Part of the UI epic #103; builds on the HPV community routing from #106. Rebased onto `main` after the evidence-map stack (#123 / #128 / #129) landed.

## What this does

Generalises the search result cards, the Refine drawer, and the evidence-map config panel so the HPV community renders correctly against its own data shape, with no change to Education (ESEA). Each difference is expressed as per-community config rather than a slug check in the components.

**Result card tags (the #125 core)**
- References surface the concepts coded against them as tags, sourced from the reference's applied-concept data (`investigationParser` gains an `appliedConcepts` slot) rather than aggregated from Education's scheme-specific fields.
- A new `inScheme` map on the vocabulary service resolves each concept to its scheme (stored as full URIs, consistent with `schemes[].uri`), which drives the geo exclusion below.

**No findings/estimates badges for HPV**
- `CommunityFeatures.findingsAndEstimates` (ESEA `true` / HPV `false`) gates the stat-badge render. HPV references carry no findings or effect estimates, so the labels would be empty.

**Geo concepts excluded from HPV pills**
- `Community.pillExcludedSchemes` drops HPV's five geography schemes from the card pills, resolved via the `inScheme` map. Kept distinct from `filterExcludedSchemes`: geo stays filterable in the drawer, it just isn't a pill.

**HPV evidence map enabled with default axes**
- `Community.defaultEvidenceMapAxes` opens HPV's map on TargetPopulation (rows) by DeliveryActor (columns). DeliveryPlatform was rejected as an axis: no coded records, so it would plot an empty map.

**Country filter hidden for HPV, in both the drawer and the map panel**
- `CommunityFeatures.countryFacetFilter` (ESEA `true` / HPV `false`) gates the typeahead Country card backed by the `countries` search facet. HPV codes country as an ISO suffix on a concept URI rather than the object property ESEA uses, so that facet is empty and the card would render "No countries match." The gate is threaded as `showCountryFacetFilter` through the shared `FilterCardList` (used by both the Refine drawer and the evidence-map config panel), so the empty card is suppressed in both places. This is a separate lever from `filterExcludedSchemes`, which only filters the vocabulary concept-scheme cards, not the hardcoded Country `FilterCard`.

**Export to Excel hidden for HPV, and off by default**
- `CommunityFeatures.exportExcel` gates the toolbar button. The default is now `false`, with ESEA opting in explicitly. Today's export runs the Education-shaped pipeline, so defaulting it off means a new community cannot silently ship a malformed workbook. Re-enabled for HPV once #127 lands.

## Why community flags, not slug checks

Each difference is a boolean or config value on `CommunityFeatures` / `Community`, so adding a third community is a config edit, not a new branch in the components. Same shape as the community-config pattern introduced in #106.

## Validation performed

- Full suite green: 774 passing across 68 files; `tsc --noEmit` clean.
- Live-verified locally against a dev server (Vite on `:3000` to a local destiny-repository on `:8000`, with 6 HPV records seeded):
  - `/hpv`: cards render applied-concept tags, no stat-badges, geo schemes excluded from pills (a 52-concept reference resolved to 6 topical pills, under the 8-pill cap), Export to Excel absent, no Country card in the Refine drawer.
  - `/hpv/visualise`: the map renders on the TargetPopulation by DeliveryActor axes with no console errors; the config panel shows HPV's Country concept-scheme card and suppresses the facet-backed Country card.
  - `/esea` and `/esea/visualise`: unchanged. Stat-badges with real counts, document-type-led pills, Export present, Country card populated with facet counts, map opens on its EducationLevel by EducationTheme defaults.
- Confirmed against the live HPV vocabulary that geo exclusion holds: `skos:inScheme` is serialised as full URIs there, matching the full-URI exclusion list.

## Follow-ups

- `countryFacetFilter` flips back to `true` for HPV as a fast-follow if/when a country projection populates the `countries` facet (not built yet). That flip also adds HPV's Country scheme to `filterExcludedSchemes` so the concept-scheme duplicate does not surface (two levers, opposite directions).
- Export stays hidden until the HPV Excel pipeline in #127 lands.

### DIFF
--- a/src/components/filters/FilterDrawer.tsx
+++ b/src/components/filters/FilterDrawer.tsx
@@ -12,7 +12,7 @@ interface FilterDrawerProps {
   open: boolean;
   title?: string;
   countNoun?: string;
-  // Show the facet-backed country card; off where the `countries` facet is empty (HPV).
+  // Show the facet-backed country card; off where the `countries` facet is empty.
   showCountryFacetFilter?: boolean;
   schemes: ConceptScheme[];
   appliedConceptFilters: readonly (readonly string[])[];
--- a/src/components/filters/FilterDrawer.tsx
+++ b/src/components/filters/FilterDrawer.tsx
@@ -12,6 +12,8 @@ interface FilterDrawerProps {
   open: boolean;
   title?: string;
   countNoun?: string;
+  // Show the facet-backed country card; off where the `countries` facet is empty (HPV).
+  showCountryFacetFilter?: boolean;
   schemes: ConceptScheme[];
   appliedConceptFilters: readonly (readonly string[])[];
   appliedCountryCodes: readonly string[];
@@ -35,6 +37,7 @@ type FilterDrawerPanelProps = Omit<FilterDrawerProps, "open">;
 function FilterDrawerPanel({
   title = "Refine the evidence",
   countNoun = "results",
+  showCountryFacetFilter = true,
   schemes,
   appliedConceptFilters,
   appliedCountryCodes,
@@ -119,7 +122,11 @@ function FilterDrawerPanel({
         </header>
 
         <div class="filter-drawer__body">
-          <FilterCardList draft={draft} countNoun={countNoun} />
+          <FilterCardList
+            draft={draft}
+            countNoun={countNoun}
+            showCountryFacetFilter={showCountryFacetFilter}
+          />
         </div>
 
         <FilterActions
--- a/src/components/search/ResultRow.tsx
+++ b/src/components/search/ResultRow.tsx
@@ -24,10 +24,16 @@ interface ResultRowProps {
   communitySlug: string;
   reference: Reference;
   codingInstitution?: CodingInstitutionConfig;
+  // ESEA shows finding/estimate stat-badges; HPV (false) hides them.
+  findingsAndEstimates?: boolean;
+  // Concept schemes whose members are dropped from pills (HPV geo).
+  pillExcludedSchemes?: readonly string[];
 }
 
 const MAX_AUTHORS_SHOWN = 3;
 export const PILL_CAP = 8;
+// Stable default so the pill useMemo isn't busted by a fresh [] each render.
+const NO_PILL_EXCLUSIONS: readonly string[] = [];
 
 function formatAuthors(authors: { display_name: string }[]): string {
   if (authors.length === 0) return "";
@@ -70,6 +76,8 @@ export function ResultRow({
   communitySlug,
   reference,
   codingInstitution: codingConfig,
+  findingsAndEstimates = true,
+  pillExcludedSchemes = NO_PILL_EXCLUSIONS,
 }: ResultRowProps) {
   const bib = extractBibliographic(reference);
   const doi = extractDoi(reference.identifiers);
@@ -80,7 +88,9 @@ export function ResultRow({
   const rawContext = linkedData?.data?.["@context"];
   const contextUrl = typeof rawContext === "string" ? rawContext : undefined;
 
-  const { labels, definitions } = useVocabulary(linkedData?.vocabulary_uri);
+  const { labels, definitions, inScheme } = useVocabulary(
+    linkedData?.vocabulary_uri,
+  );
   const { context } = useContextPrefixes(contextUrl);
 
   const pillTags = useMemo(() => {
@@ -90,7 +100,17 @@ export function ResultRow({
       context.prefixes,
       labels,
     );
-    const concepts = aggregatePillConcepts(investigation);
+    // HPV carries flat hasAppliedConcept; ESEA builds pills from the
+    // documentType/findings walk. The two shapes are disjoint.
+    const sourced =
+      investigation.appliedConcepts.length > 0
+        ? investigation.appliedConcepts.map((value) => ({ value }))
+        : aggregatePillConcepts(investigation);
+    const excluded = new Set(pillExcludedSchemes);
+    const concepts =
+      excluded.size === 0
+        ? sourced
+        : sourced.filter((a) => !excluded.has(inScheme?.get(a.value.uri) ?? ""));
     // Flat pills (no parent breadcrumb) — pass an empty `broader` map.
     // Definitions still drive the hover tooltip.
     return conceptsToTags(
@@ -99,7 +119,7 @@ export function ResultRow({
       new Map(),
       definitions ?? new Map(),
     );
-  }, [linkedData, labels, definitions, context]);
+  }, [linkedData, labels, definitions, inScheme, context, pillExcludedSchemes]);
 
   const title = bib?.title ?? reference.id;
   const authors = bib?.authorship ? formatAuthors(bib.authorship) : "";
@@ -161,12 +181,16 @@ export function ResultRow({
             DOI ↗
           </a>
         )}
-        <span class="stat-badge">
-          <span class="stat-num">{findingsLabel}</span> findings
-        </span>
-        <span class="stat-badge">
-          <span class="stat-num">{estimatesLabel}</span> estimates
-        </span>
+        {findingsAndEstimates && (
+          <>
+            <span class="stat-badge">
+              <span class="stat-num">{findingsLabel}</span> findings
+            </span>
+            <span class="stat-badge">
+              <span class="stat-num">{estimatesLabel}</span> estimates
+            </span>
+          </>
+        )}
         {codingInstitution && (
           <span class="row-coder" data-testid="coder-text">
             {codingInstitution}
--- a/src/components/search/ResultRow.tsx
+++ b/src/components/search/ResultRow.tsx
@@ -24,9 +24,9 @@ interface ResultRowProps {
   communitySlug: string;
   reference: Reference;
   codingInstitution?: CodingInstitutionConfig;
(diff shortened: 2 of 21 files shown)


src/components/filters/FilterDrawer.tsx
```diff
@@ -42,6 +42,8 @@ interface FilterDrawerProps {
   open: boolean;
   title?: string;
   countNoun?: string;
+  // Show the facet-backed country card; off where the `countries` facet is empty (HPV).
```
jack-o-the-green: `nitpick` remove `(HPV)` to keep concrete community references out of the FilterDrawer

src/components/search/ResultRow.tsx
```diff
@@ -24,10 +24,16 @@ interface ResultRowProps {
   communitySlug: string;
   reference: Reference;
   codingInstitution?: CodingInstitutionConfig;
+  // ESEA shows finding/estimate stat-badges; HPV (false) hides them.
```
jack-o-the-green: `nitpick` same as above, remove community specific references from components where possible. This is the sort of thing we'd likely forget to update if/when the default for HPV or other communities change, resulting in inconsistent comments.

src/pages/SearchPage.tsx
```diff
@@ -335,7 +335,7 @@ function SearchPageInner({ community }: { community: Community }) {
                     {exportAnnouncement}
                   </span>
                 )}
-                {results.results && (
+                {results.results && community.features.exportExcel && (
```
jack-o-the-green, line 359: This community specific feature flagging is really nice, it's very readable!

src/services/communities.ts
```diff
@@ -17,8 +17,21 @@ function requireEnv(name: string, value: string | undefined): string {
 
 const DEFAULT_FEATURES: CommunityFeatures = {
   evidenceMap: false,
+  findingsAndEstimates: true,
+  exportExcel: true,
```
jack-o-the-green: Let's default `exportExcel` to `false` since excel exports are currently hardcoded for a single community - less change of accidentally having it on later.

src/services/communities.ts
```diff
@@ -17,8 +17,21 @@ function requireEnv(name: string, value: string | undefined): string {
...
+  countryFacetFilter: true,
 };
 
+// HPV geo ConceptSchemes dropped from result-card pills (#125): they swamp the
+// pill cap and geo has dedicated filters. Full scheme URIs to match inScheme.
+const HPV_GEO_SCHEMES = [
```
jack-o-the-green, line 28: Possibly too much scope creep, but I was wondering if an alternative approach would be preferable here. 

We could
* Set a priority order for concept schemes or just treat this list as "low priority"
* Set a "max number of pills displayed" and then simply render the "highest priorty" pills with a `[+]` if there are more than the maximum number of pills in the concept list. 

That way if we only get geo information for a reference, we still display as much as we can to the user - otherwise a reference with only geo schemes coded would have empty pills. 

Feel free to ignore this option, it may be more effort than we want to go to at this stage. Perhaps better as a follow up but keen to get your thoughts!

jack-o-the-green (commented):
This is looking really good - super tidy implementation. 

Thinking about merge priority on the overlap between this and #128  as called out in the description - we've expressed desire to get the evidence map in for jacobs tomorrow, so am thinking it's best to hold off merging this until we get the other PR chain in. Feel free to disagree with me!

jaybea:
Totally agree to hold off merge (and even starting improvements from your comments) until after Jacobs demo prep for Tuesday!

src/components/filters/FilterDrawer.tsx
jaybea: Done.

src/components/search/ResultRow.tsx
jaybea: Done.

src/services/communities.ts
jaybea: Done.

src/services/communities.ts
jaybea: I was thinking about this too, thanks for surfacing it. But short-term let's keep excluding geo from the pills. In the Mansilla paper case, it has 52 applied concepts, 46 of them geo, so it would render the 8 pills, then say +44 more. Kinda feel geo in pills doesn't add so much.

src/components/filters/FilterDrawer.tsx
jack-o-the-green: @jaybea I don't know if my diff is wack for some reason but I'm still seeing the `(HPV)` here

src/components/search/ResultRow.tsx
```diff
@@ -90,7 +100,17 @@ export function ResultRow({
       context.prefixes,
       labels,
     );
-    const concepts = aggregatePillConcepts(investigation);
+    // HPV carries flat hasAppliedConcept; ESEA builds pills from the
```
jack-o-the-green: More community specific information in components

src/types/models.ts
```diff
@@ -1,6 +1,12 @@
 export interface CommunityFeatures {
   // No UI yet; reserved so the evidence map can ship behind a flag (#103).
   evidenceMap: boolean;
+  // ESEA studies carry findings/estimates (the stat-badges); HPV references don't.
+  findingsAndEstimates: boolean;
+  // HPV export runs the Education-shaped pipeline today; hide it until #127.
```
jack-o-the-green: All this documentation for the ESEA/HPV communities will become out of date without requiring changes to this file (because it's just a model, actual changes happen in each community). I know I'm being really picky about these (sorry!) but it's a way that we will get confused later because our documentation will become inconsistent.

jack-o-the-green (approved):
Approved pending documentation nitpicks

jaybea:
I did miss the community specific doc-cleansing, pretty sloppy! 
Thanks for catching it @jack-o-the-green .

PR #142, opened 2026-06-10
Add landing page for unauthenticated hpv community users

Adam-Hammo:
Adds a landing page for anyone that accesses `/hpv` (or its sub-paths) without being logged in. The login button takes them to the normal flow, `Create Account` goes directly to the keycloak registration form.

`/esea` or `/` autodirect to the keycloak login page (existing behaviour).

image

### DIFF
--- a/public/silent-check-sso.html
+++ b/public/silent-check-sso.html
@@ -1,0 +1,8 @@
+<!doctype html>
+<html>
+  <body>
+    <script>
+      parent.postMessage(location.href, location.origin);
+    </script>
+  </body>
+</html>
--- a/src/auth/AuthGate.tsx
+++ b/src/auth/AuthGate.tsx
@@ -1,3 +1,27 @@
+import { login, register } from "./keycloak";
+
+export function Landing() {
+  return (
+    <div class="auth-gate">
+      <div class="auth-gate__panel">
+        <p>Welcome to the Evidence Repository.</p>
+        <div class="auth-gate__actions">
+          <button
+            type="button"
+            class="auth-gate__button auth-gate__button--primary"
+            onClick={login}
+          >
+            Sign in
+          </button>
+          <button type="button" class="auth-gate__button" onClick={register}>
+            Create account
+          </button>
+        </div>
+      </div>
+    </div>
+  );
+}
+
 export function Loading() {
   return (
     <div class="auth-gate">
--- a/src/auth/keycloak.ts
+++ b/src/auth/keycloak.ts
@@ -1,5 +1,6 @@
 import Keycloak from "keycloak-js";
 import { KEYCLOAK_CLIENT_ID, KEYCLOAK_REALM, KEYCLOAK_URL } from "@/config";
+import { findCommunity } from "@/services/communities";
 
 function requireEnv(name: string, value: string | undefined): string {
   if (!value) {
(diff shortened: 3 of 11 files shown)


jaybea (approved):
Nice - works as described, and the per-community gate is the right shape. Approve after a few should-fixes, none blocking.

**1. `check-sso` bounces through Keycloak before the landing renders (UX).** With `onLoad: "check-sso"`, `checkLoginIframe: false`, and no `silentCheckSsoRedirectUri`, keycloak-js does a full-page `prompt=none` redirect to Keycloak and back before `<Landing/>` paints (verified against keycloak-js 26.2.4: it falls to `login({prompt:"none"})`). So a new unauthenticated HPV visitor bounces through Keycloak before seeing the Sign in / Create account screen, on every anonymous load. For a page meant to be the local front door, that round-trip is unexpected. To be clear, this isn't a resilience argument - we host Keycloak, so if it's down the whole app is down anyway; it's purely the happy-path flash.

Options:

| Option | Bounce on anon load? | Auto-detect existing session? | Cost |
|---|---|---|---|
| `check-sso`, no silent URI (today) | Always | (the bounce is the check) | flash on every anon load |
| `check-sso` + silent URI, default fallback | Restrictive browsers only | capable: yes; restrictive: bounce | needs `silent-check-sso.html`; inconsistent |
| `check-sso` + silent URI + `silentCheckSsoFallback: false` | Never | capable: yes; restrictive: no | needs `silent-check-sso.html` |
| omit `onLoad` for self-signup routes | Never | No (everyone clicks Sign in once) | one line |

The last two differ only in whether a returning signed-in user is auto-logged-in on a capable browser; on a restrictive/cookie-blocked browser they're identical (landing + one click). For an open-access front door that's mostly new users, I'd take the one-liner:

```ts
onLoad: allowsSelfSignup() ? undefined : "login-required",
```

`init` still processes the `?code=` callback on the way back from `login()`, so Sign in / Create account work unchanged; the only loss is auto-login on landing (returning users click Sign in once, no credential prompt). If returning-user auto-login matters, the proper path is `silentCheckSsoRedirectUri` + `silentCheckSsoFallback: false` instead. Either's fine - take it or leave it.

**2. `SELF_SIGNUP_COMMUNITIES = new Set(["hpv"])` is a second source of truth.** We already model per-community capability in `CommunityFeatures` off the `COMMUNITIES` registry, and ESEA-no / HPV-yes self-signup is exactly that kind of policy. Consider a `selfSignup` flag and `findCommunity(slug)?.features.selfSignup ?? false` - one registry edit to add a community instead of a parallel list, and it drops the slug re-parse (the router already extracts `:community`). `communities.ts` is already imported at startup, so no new failure surface.

**3. Tests.** `AuthGate.test.tsx` covers `Loading`/`AuthError` in the exact shape a `Landing` test would take - worth adding that (trivial). The slug→`onLoad` decision is the branch most worth pinning since it breaks silently when a community is added, though that one needs a `keycloak.init` mock.

**Nit:** `.auth-gate__button` duplicates `.auth-gate__retry` verbatim (incl. hover) - could share a base.

Validated locally: anon `/hpv` renders the landing with zero Keycloak requests; Sign in builds a correct auth-code + PKCE request with the `/hpv` redirect preserved.

Adam-Hammo:
Thanks @jaybea , actioned

On point 1, full removal doesn't work as then the `hpv` page doesn't know if you're signed in or not. I've replaced it with a silent check instead of the redirect flow.

jaybea (commented):
Looks good!

jaybea (approved):
Looks good!

PR #146, opened 2026-06-12
HPV reference detail: Taxonomy codes card (+ HPV map axis defaults and Countries-axis gate)

### LINKED ISSUE #126: HPV reference detail: split bibliographic and Taxonomy codes cards
Part of the UI epic [#103](https://github.com/destiny-evidence/evidence-repo-ui/issues/103). Builds on [#106](https://github.com/destiny-evidence/evidence-repo-ui/issues/106) and [#125](https://github.com/destiny-evidence/evidence-repo-ui/issues/125).

## User story

Clicking a reference on `/hpv` opens a detail page with two visually distinct cards:

1. A **bibliographic card** — title, authors, journal, DOI, abstract.
2. A **"Taxonomy codes" card** below, listing the applied concepts grouped by scheme. Where a scheme has sub-categories (e.g. *Country or regional focus* with Continent / World Bank Region / Income Group), they render as sub-headings within that block.

What's different from Education's reference detail:

- **No Findings section** — HPV references don't have findings.
- **A new "Taxonomy codes" card** replaces the scheme-specific Education sections.

## What stays the same

- Education (`/esea`) reference detail behaves exactly as today.
- Card chrome, spacing, and typography match the existing design tokens.

## Verify

- HPV reference detail layout matches the `03 Reference detail` mockup in the HPV design.
- `/esea` reference detail unchanged; existing tests pass.

## Pointers

- Reference detail page lives in [`src/pages/RecordDetailPage.tsx`](src/pages/RecordDetailPage.tsx).
- The two cards share chrome, so a small reusable `<Card>` primitive is worth extracting (props: `header`, `children`).
- Sub-grouping within a scheme can use `skos:broader` from the loaded vocab.
- HPV applied-concept data flows in via [#125](https://github.com/destiny-evidence/evidence-repo-ui/issues/125) — reuse the same field rather than re-parsing.

## Depends on

- [#125](https://github.com/destiny-evidence/evidence-repo-ui/issues/125) for the applied-concept data path.
- End-to-end verification depends on [destiny-repository#731](https://github.com/destiny-evidence/destiny-repository/issues/731).

jaybea:
Closes #126.

## Summary

Adds a **Taxonomy codes** card to the `/hpv` reference detail page, below the bibliographic card. It lists every applied concept grouped by concept scheme, geo schemes first, with each scheme shown in full. ESEA reference detail is unchanged.

The branch also folds in two small HPV evidence-map changes that surfaced while enabling the HPV community, cherry-picked from the map fix work and verified together:
- HPV evidence-map default axes changed to WHO Region (rows) x Thematic Focus - Primary (columns), per the PM's directive, replacing the earlier Target Population x Delivery Actor placeholder.
- The map config panel now hides the "Countries" axis option when a community's country facet is empty (HPV). Previously it was offered unconditionally; selecting it ran a cross-facet query against a missing facet and failed the whole map. Mirrors the existing country filter-card gate.

## What changed

Reference detail card (the issue):
- New pure transform `groupAppliedConcepts` (`src/services/taxonomyCodesUtils.ts`): buckets applied concepts by `inScheme`, prunes each scheme's `topConcepts` tree to the branches that reach a coded concept (ancestors kept as sub-headings, arbitrary depth), and orders geo schemes first then topical (published-vocab order within each). Concepts whose scheme is unknown fall to a trailing "Other codes" bucket.
- New presentational `TaxonomyCodesCard` (`src/components/investigation/TaxonomyCodesCard.tsx`): renders `.lg-card` inline (no shared Card primitive), one section per scheme listing all its members as tags via the existing `TagGroup`.
- Wired into `RecordDetailPage`, data-gated on `appliedConcepts.length > 0`.

Supporting:
- Consolidated `Community.pillExcludedSchemes` into a single `Community.geographicSchemes` field. One source of truth for "what is geo": it drives both result-card pill exclusion (at the SearchPage seam) and detail-page ordering, so the two cannot drift.
- Split the page's label-failure flag in two: `vocabUnavailable` (vocab fetch only) for the HPV-only Taxonomy codes card, and `curieLabelsUnavailable` (vocab or context) for the InvestigationCard and findings banner. HPV concepts are full URIs and resolve from the vocab alone; ESEA concepts are CURIEs that also need the `@context` to expand the prefix, so a context-only failure must not trip the HPV card's raw-identifier note.

## Design notes

- **Data-gated, not community-gated (deliberate).** The card renders for any community with applied concepts. ESEA emits none today (it builds tags from its findings walk, a disjoint data path), so ESEA never shows the card. This keeps the gate data-driven and mirrors how the result-card pills already work, rather than hard-coding an HPV slug check. An integration test pins this: an ESEA record that did carry applied concepts would get the card.
- **Every concept is shown in full (no roll-up).** Country coding can flood (one gold-set ref carries 83 countries). An earlier revision compacted any scheme past 10 codes to an expand-on-click "Multiple {plural} (N)" summary; the team reversed that call: they want all codes shown, and even 25 countries reads as about three wrapped lines. The summary path and its threshold were removed, so each scheme lists every member, however many. The grouping transform already built the full node tree, so this was a presentation-only deletion.
- **Ordering follows published-vocab order** within the geo-first / topical split. It is not yet aligned to the Refine drawer's ordering.
- **Sub-category nesting is kept vocab-faithful.** Where the vocab asserts `broader` (e.g. World Bank income tier and Gavi status under Country Classification), the card renders the ancestor as a sub-heading. The mockup drew these flat; keeping them nested is a deliberate fidelity-to-vocab choice, flagged for design awareness.

## Validation performed

- Full suite green: 800 tests across 71 files; `tsc --noEmit` clean.
- Live-verified locally (dev server on `:3000` to a local destiny-repository on `:8000`, 6 seeded HPV records):
  - All 6 HPV reference-detail pages render the Taxonomy codes card; the 133 applied-concept URIs across the 6 records resolve to readable labels with zero misses (country codes show as full names, e.g. "Afghanistan", not "AF").
  - HPV evidence map: the WHO Region x Thematic Focus - Primary defaults render populated bubbles (African Region, European Region, Region of the Americas) against staging data; the Countries axis option is correctly absent for HPV, so the map no longer fails on that selection.
- Live-verified against the staging backend on genuinely robot-mapped HPV references (not just local fixtures): a mapped record renders the card correctly, geo schemes first (Country Classification, Country), with multi-level vocab nesting (e.g. World Bank income tier under Country Classification). The Octavilia literature review codes 12 countries and lists all 12 inline with no summary toggle, confirming the roll-up removal; `read_page` finds no disclosure control on the page and the console is clean. Un-mapped references correctly show no card (bibliographic only).
- ESEA reference detail is unaffected by design (the card is data-gated on applied concepts, which ESEA does not emit) and is covered by integration tests; the ESEA search and visualise pages were confirmed unchanged during the HPV map enablement.

### DIFF
--- a/src/components/investigation/TaxonomyCodesCard.tsx
+++ b/src/components/investigation/TaxonomyCodesCard.tsx
@@ -45,22 +45,27 @@ function Nodes({ nodes }: { nodes: TaxoNode[] }) {
   );
 }
 
+// "Country" → "countries". Lower-cased plural for the roll-up summary pill.
+// Already-plural / sibilant labels are left alone; only the geo Country scheme
+// floods past the threshold today, so the y→ies case is the one that matters.
+function pluralLower(label: string): string {
+  const lower = label.toLowerCase();
+  if (lower.endsWith("s")) return lower;
+  if (lower.endsWith("y")) return `${lower.slice(0, -1)}ies`;
+  return `${lower}s`;
+}
+
 function SchemeBlock({ group }: { group: TaxonomyGroup }) {
-  const body = <Nodes nodes={group.nodes} />;
-  if (group.collapsedByDefault) {
-    return (
-      <details class="taxonomy-codes-card__group">
-        <summary class="taxonomy-codes-card__scheme">
-          {group.schemeLabel} ({group.appliedCount})
-        </summary>
-        {body}
-      </details>
-    );
-  }
   return (
     <section class="taxonomy-codes-card__group">
-      <h3 class="taxonomy-codes-card__scheme">{group.schemeLabel}</h3>
-      {body}
+      <h3 class="taxonomy-codes-card__scheme lg-section-label">
+        {group.schemeLabel}
+      </h3>
+      {group.rolledUp ? (
+        <TagGroup tags={[{ label: `Multiple ${pluralLower(group.schemeLabel)}` }]} />
+      ) : (
+        <Nodes nodes={group.nodes} />
+      )}
     </section>
   );
 }
@@ -72,7 +77,7 @@ export function TaxonomyCodesCard({
   if (groups.length === 0) return null;
   return (
     <article class="taxonomy-codes-card lg-card">
-      <h2 class="taxonomy-codes-card__title">Taxonomy codes</h2>
+      <h2 class="taxonomy-codes-card__title lg-kicker">Taxonomy codes</h2>
       {vocabUnavailable && (
         <p class="taxonomy-codes-card__vocab-note" role="status">
           Vocabulary unavailable — codes are shown as raw identifiers.
--- a/src/components/investigation/TaxonomyCodesCard.tsx
+++ b/src/components/investigation/TaxonomyCodesCard.tsx
@@ -8,41 +8,19 @@ interface TaxonomyCodesCardProps {
   vocabUnavailable?: boolean;
 }
 
-// Split one nesting level into chips (applied concepts) and sub-heading blocks
-// (non-applied ancestors). An applied node's own children are hoisted up to its
-// level, so a coded parent and its coded child both render — neither is dropped.
-function partition(nodes: readonly TaxoNode[]): {
-  tags: HierarchicalTag[];
-  subheadings: TaxoNode[];
-} {
+// Flatten a scheme's pruned tree into one tag per coded concept, carrying its
+// immediate ancestor's label as a breadcrumb (rendered "Parent › Child" inline,
+// matching the findings cards). Non-applied ancestors are not tags themselves;
+// they only supply the breadcrumb for their coded descendants.
+function flatten(nodes: readonly TaxoNode[], parent?: string): HierarchicalTag[] {
   const tags: HierarchicalTag[] = [];
-  const subheadings: TaxoNode[] = [];
   for (const n of nodes) {
     if (n.applied) {
-      tags.push({ label: n.label, definition: n.definition });
-      const hoisted = partition(n.children);
-      tags.push(...hoisted.tags);
-      subheadings.push(...hoisted.subheadings);
-    } else {
-      subheadings.push(n);
+      tags.push({ label: n.label, definition: n.definition, parent });
     }
+    tags.push(...flatten(n.children, n.label));
   }
-  return { tags, subheadings };
-}
-
-function Nodes({ nodes }: { nodes: TaxoNode[] }) {
-  const { tags, subheadings } = partition(nodes);
-  return (
-    <>
-      {tags.length > 0 && <TagGroup tags={tags} />}
-      {subheadings.map((a) => (
-        <div key={a.uri} class="taxonomy-codes-card__subgroup">
-          <h4 class="taxonomy-codes-card__subheading">{a.label}</h4>
-          <Nodes nodes={a.children} />
-        </div>
-      ))}
-    </>
-  );
+  return tags;
 }
 
 function SchemeBlock({ group }: { group: TaxonomyGroup }) {
@@ -51,7 +29,7 @@ function SchemeBlock({ group }: { group: TaxonomyGroup }) {
       <h3 class="taxonomy-codes-card__scheme lg-section-label">
         {group.schemeLabel}
       </h3>
-      <Nodes nodes={group.nodes} />
+      <TagGroup tags={flatten(group.nodes)} />
     </section>
   );
 }
--- a/src/components/investigation/TaxonomyCodesCard.tsx
+++ b/src/components/investigation/TaxonomyCodesCard.tsx
@@ -1,7 +1,5 @@
-import { useId, useState } from "preact/hooks";
 import { TagGroup } from "../common/TagGroup";
 import type { HierarchicalTag } from "../common/TagGroup";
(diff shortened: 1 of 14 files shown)


src/pages/RecordDetailPage.tsx
```diff
@@ -112,11 +139,17 @@ export function RecordDetailPage({ id }: RecordDetailPageProps) {
           codingInstitution={codingInstitution}
           isRetracted={isRetracted}
           hasInvestigation={hasLinkedData}
-          vocabUnavailable={!!vocabUnavailable}
+          vocabUnavailable={curieLabelsUnavailable}
         />
+        {taxonomyGroups.length > 0 && (
```
jack-o-the-green, line 144: The padding around this card is quite different to the findings section, let's add the padding so it matches the wireframe better. 

Image
Image

src/components/investigation/TaxonomyCodesCard.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+interface TaxonomyCodesCardProps {
+  groups: TaxonomyGroup[];
+  vocabUnavailable?: boolean;
+}
+
+// Split one nesting level into chips (applied concepts) and sub-heading blocks
+// (non-applied ancestors). An applied node's own children are hoisted up to its
+// level, so a coded parent and its coded child both render — neither is dropped.
+function partition(nodes: readonly TaxoNode[]): {
```
jack-o-the-green: Instead of partitioning, could we match the findings cards with the inlined hierarchy? It would make the total length of the card shorter and remove the indented pills which look quite messy to my eyes. 

Image

src/services/taxonomyCodesUtils.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export interface TaxonomyGroup {
+  schemeUri: string;
+  schemeLabel: string;
+  // drives ordering only (geo first)
+  isGeo: boolean;
+  nodes: TaxoNode[];
+}
+
+const OTHER_SCHEME_URI = "__other_codes__";
```
jack-o-the-green, line 26: Asking for mapping quality reasons - did you find any records that had schemes not in the taxonomy or is this for future resilience?

src/types/models.ts
```diff
@@ -55,9 +53,10 @@ export interface Community {
   vocabularyUrl: string;
   contextUrl: string;
   filterExcludedSchemes: string[];
-  // Concept schemes dropped from result-card pills (e.g. geo). Distinct from
-  // filterExcludedSchemes (drawer): geo stays filterable, just isn't a pill.
-  pillExcludedSchemes: string[];
+  // Geographic concept schemes (country + regional/classification): excluded from
+  // result-card pills and deprioritized on the detail page. Distinct from
```
jack-o-the-green: I think these are prioritized on the detail page now as they're at the top

src/pages/SearchPage.tsx
```diff
@@ -461,7 +410,7 @@ function SearchPageInner({ community }: { community: Community }) {
               reference={ref}
               codingInstitution={community.codingInstitution}
               findingsAndEstimates={community.features.findingsAndEstimates}
-              pillExcludedSchemes={community.pillExcludedSchemes}
+              pillExcludedSchemes={community.geographicSchemes}
```
jack-o-the-green: I feel a little sad that we're losing the ability to designate schemes for pill exclusion at the community level, it's less obvious to me that setting geographicSchemes in a community would result in them being removed from pills. 

Wondering if we can keep the `pillExcludedSchemes` in community config and simply add the geographic schemes to it for. hpv

jack-o-the-green (commented):
This is looking good! I like the choice to have the card show up whenever applied concepts are populated, feels like a great way to avoid tying to community. 

Functionality looks great - couple comments on how we could make things look more like the wireframe and a request to retain `pillExcludedSchemes` on the community model.

src/components/investigation/TaxonomyCodesCard.tsx
jack-o-the-green: Happy for this one to happen in a follow up if it's tricky to do - I know we want to get this out today.

jaybea: I love these!

src/pages/RecordDetailPage.tsx
jaybea: Found it! Missing margin top.

src/services/taxonomyCodesUtils.ts
jaybea: Future resilience. Man was this hard to verify "live". Had to proxy the vite server to feed it a fiddled vocabulary!

src/types/models.ts
jaybea: You're right, stale comment.

src/pages/SearchPage.tsx
jaybea: They baaack! Good call and it clarifies the meaning. I was going minimal and pushed it too far.

src/services/taxonomyCodesUtils.ts
jack-o-the-green: Phew - that's good to know! Good resilience to have too

jack-o-the-green (approved):
LGTM 

Looks so so good 🎉

PR #151, opened 2026-06-13
Correctly order schemes in filter drawer and concepts in evidence map

### LINKED ISSUE #148: Evidence maps: update how parent/children concepts appear in the evidence map
right now the rows appear in alphabetical order - which means parents and children aren't nested or appear next to one another

Image

### LINKED ISSUE #149: Reorder filters for better usability
1. First: learning convening
2. Second: year
3. Next: anything geo related (country, various regions)
4. After: rest of the filters in alpha-numerical order

Within a concept scheme:
- alpha-numeral order

Adam-Hammo:
Closes #148 - orders evidence map concepts by hierarchy level first, so parents always appear before children. Then ordered alphabetically.
Before & after:
image
image

Closes #149 - adds `pinnedFilters` to communities that display those filters first in given order, before the rest alphabetically.
Before & after:
image
image

### DIFF
--- a/src/components/filters/filterOrder.ts
+++ b/src/components/filters/filterOrder.ts
@@ -1,0 +1,89 @@
+import type { FilterSlot } from "@/types/models";
+import {
+  schemeDisplayLabel,
+  type ConceptScheme,
+} from "@/services/vocabulary/vocabularyService";
+
+export type { FilterSlot };
+
+// One filter card, in render order: the two built-in cards plus one per scheme.
+export type FilterItem =
+  | { kind: "year" }
+  | { kind: "country" }
+  | { kind: "scheme"; scheme: ConceptScheme };
+
+// Year first, then the country facet, then every scheme alphabetically.
+export const DEFAULT_FILTER_ORDER: readonly FilterSlot[] = [
+  "year",
+  "country",
+  "otherSchemes",
+];
+
+interface OrderFilterItemsOptions {
+  order?: readonly FilterSlot[];
+  // Scheme URIs grouped by the "geographicSchemes" slot, in this order.
+  geographicSchemes?: readonly string[];
+  // When false the "country" slot renders nothing (facet unavailable).
+  showCountryFacetFilter?: boolean;
+}
+
+/**
+ * Resolve a community's {@link FilterSlot} order into the concrete list of
+ * filter cards to render (#149). Each scheme appears exactly once: a slot naming
+ * its URI, or the "geographicSchemes"/"otherSchemes" group that claims it first,
+ * wins. A trailing "otherSchemes" sweep is always applied so no scheme is
+ * silently dropped, even if the configured order omits it.
+ */
+export function orderFilterItems(
+  schemes: readonly ConceptScheme[],
+  {
+    order = DEFAULT_FILTER_ORDER,
+    geographicSchemes = [],
+    showCountryFacetFilter = true,
+  }: OrderFilterItemsOptions = {},
+): FilterItem[] {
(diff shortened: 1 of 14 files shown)


src/components/filters/filterOrder.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (slot === "year") items.push({ kind: "year" });
+    else if (slot === "country") {
+      if (showCountryFacetFilter) items.push({ kind: "country" });
+    } else pushScheme(slot);
+  }
+
+  const rest = schemes
+    .filter((s) => !placed.has(s.uri))
+    .sort((a, b) =>
+      schemeDisplayLabel(a.label).localeCompare(schemeDisplayLabel(b.label)),
```
jaybea: I think this needs a `{ numeric: true }`.

src/services/evidenceMap.ts
```diff
@@ -128,18 +130,27 @@ function flattenScheme(scheme: ConceptScheme): AxisCategory[] {
   return out;
 }
 
+const byLabelThenKey = (
+  a: { label: string; uri?: string; key?: string },
+  b: { label: string; uri?: string; key?: string },
+): number =>
+  a.label.localeCompare(b.label) ||
```
jaybea: I think this needs a `{ numeric: true }`. A test flushing this out would be really good too.

src/components/filters/filterOrder.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  const pushScheme = (uri: string) => {
+    const scheme = byUri.get(uri);
+    if (scheme && !placed.has(uri)) {
+      placed.add(uri);
+      items.push({ kind: "scheme", scheme });
+    }
+  };
+
+  for (const slot of pinned) {
```
jaybea, line 49: Belt and suspenders, maybe dedup on adding year and country (check and only add if not present already)? It would be a config error but just feels better since pushing to a list to dedup.

src/services/evidenceMap.ts
```diff
@@ -128,18 +130,27 @@ function flattenScheme(scheme: ConceptScheme): AxisCategory[] {
...
+  (a.uri ?? a.key ?? "").localeCompare(b.uri ?? b.key ?? "");
+
+// Axis categories keep their given order — a scheme's hierarchy, or the empty
+// list for a countries axis. Cell-only keys the axis doesn't enumerate trail
+// them, alphabetized; a countries axis (no categories) is thus fully alphabetical.
```
jaybea, line 141: Nit-picky: All catergories used to go through a Map, so were deduplicated by key (the concept URL). None of our data has a duplicate currently, but a creative vocab designer (arguably malformed) could have a top concept *and* give it a `broader` into another subtree. The same URI then gets emitted twice and React would issue a duplicate-key warning.

Or maybe this (very hypothetical at this point) name collision needs to be fixed at the `flattenScheme` level.

jaybea (changes requested):
This looks good, and I'll bet user of this will appreciate the usability of this ordering.

And I think Jack will especially will appreciate the alphabetising!

But:
1. We missed the 2nd half of Issue #149:
    Within a concept scheme:
         alpha-numeral order

I couldn't find where the concepts are sorted within a scheme in the code. They almost always ARE in the vocabs we've seen. But here's one Target Population: 

Image

2. Also the `{ numeric: true }`:

Image

I think these 2 should be fixed up before we call sorted done (unless I'm missing something).

Your choice on the nits.

Adam-Hammo:
Great pickups @jaybea, I can't believe I missed that! All addressed

jaybea (approved):
Nice fixes, very centralised instead of DRY.

PR #152, opened 2026-06-13
Persist AI summary runs in background

Adam-Hammo:
Simple implementation of behaviour discussed in https://covidence.slack.com/archives/C0810ARPCRG/p1781261286936989?thread_ts=1781238605.371139&cid=C0810ARPCRG

Lifts the AI summary drawer up into the main application so it can be invoked from other pages. The summary indicator persists across navigation. 

New summaries may not be requested until the existing one is either cancelled, or finished and viewed.

Weaknesses: does not persist across hard refreshes or new tabs. `Cancel` does not actually cancel the server-side request - the backend does not expose that functionality.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import Router from "preact-router";
 import { AuthProvider } from "./auth/AuthContext";
 import { CommunityProvider } from "./community/CommunityContext";
+import { AiSummaryProvider } from "./components/ai-summary/AiSummaryProvider";
 import { AppShell } from "./components/layout/AppShell";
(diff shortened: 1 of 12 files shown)


jaybea (approved):
This is cool, love the pill in the lower right. Did I just not see it before? I clicked on close while generating my 2nd ai summary, and couldn't really find it. I knew requesting the same summary would bring it up since they're cached, or persisted.

Clicking it brings up the summary nicely.

It would be nice to have integration tests that prove this new behaviour, something like:
"keeps a background summary across same-community routes and blocks a new run until viewed", and
 "drops the active summary when the user changes community"

PR #164, opened 2026-06-15
Sort by newest if sort field is relevance and there is no keyword search

Adam-Hammo:
Essentially sets `newest` as a secondary sort when relevance is unable to distinguish records.


jaybea (approved):
Looks good. Some would even say elegant!

PR #165, opened 2026-06-15
Add HPV reference-level Excel export (#127)

### LINKED ISSUE #127: HPV Export to Excel: reference-level workbook for HPV community

danielbreves:
Closes #127.

Adds a per-community workbook dispatch so `/hpv` "Export to Excel" produces a single-sheet, one-row-per-reference workbook (bibliographic columns + one column per SKOS scheme), while `/esea` keeps its three-sheet investigation workbook unchanged.

- `Community.exportVariant` (`esea` | `hpv`) selects the builder; HPV's `exportExcel` feature is now on.
- New `buildHpvRows.ts` groups each reference's applied concepts by scheme (all schemes included, including geo; one row per reference even when concept-less).
- Dev-only blob proxy (`/blob-proxy`, mirroring `/vocab-proxy`) so local dev can fetch signed export blobs — the storage CORS only allows deployed UI origins. Inert in production.

### To do:
- [x] Confirm export format

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,9 @@
 VITE_API_TARGET=http://localhost:8000
 VITE_VOCAB_PROXY_TARGET=https://vocab.staging.evidence-repository.org
+# Proxies signed search-export blob downloads in local dev: the deployed storage
+# account's CORS allows only the production UI origins, not localhost. Set to the
+# blob endpoint of the storage account the API hands back signed URLs for.
+VITE_BLOB_PROXY_TARGET=https://destinyrepositorystagsa.blob.core.windows.net
 VITE_KEYCLOAK_URL=https://auth.evidence-repository.org
 VITE_KEYCLOAK_REALM=destiny
 VITE_KEYCLOAK_CLIENT_ID=evidence-repo-ui-client-development
--- a/src/config.ts
+++ b/src/config.ts
@@ -41,3 +41,20 @@ export function proxyVocabUrl(url: string): string {
   }
   return url;
 }
+
+const BLOB_PROXY_TARGET = import.meta.env.VITE_BLOB_PROXY_TARGET;
+
+/**
+ * Rewrite a signed blob-storage URL (e.g. a search-export download) through the
+ * local dev proxy when VITE_BLOB_PROXY_TARGET is set. The deployed storage
+ * account's CORS allows only the production UI origins, so local dev can't fetch
+ * the blob cross-origin; the proxy makes it same-origin. The SAS query string is
+ * preserved. In production, returns the URL unchanged.
+ */
+export function proxyBlobUrl(url: string): string {
+  if (!BLOB_PROXY_TARGET) return url;
+  if (url.startsWith(BLOB_PROXY_TARGET)) {
+    return "/blob-proxy" + url.slice(BLOB_PROXY_TARGET.length);
+  }
+  return url;
+}
--- a/src/hooks/useSearchExport.ts
+++ b/src/hooks/useSearchExport.ts
@@ -6,7 +6,11 @@ import {
   type SearchFilters,
 } from "@/services/apiClient";
 import { exportReferencesToExcel } from "@/services/export/export";
-import type { CodingInstitutionConfig, SearchExportRead } from "@/types/models";
+import type {
+  CodingInstitutionConfig,
+  ExportVariant,
+  SearchExportRead,
+} from "@/types/models";
 
 export type ExportStatus =
   | "idle"
@@ -22,6 +26,7 @@ export interface StartExportOptions {
   filename: string;
   vocabularyUrl: string;
   contextUrl: string;
+  variant: ExportVariant;
   codingInstitution?: CodingInstitutionConfig;
 }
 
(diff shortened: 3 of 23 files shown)


jaybea (commented):
Some early suggestions. I haven't had a chance to live validate yet, these are from static analysis and just "heads up" findings.

We should probably export EPPI ItemId somewhere, currently only doing DOI, but at least 40 records had no valid external identifiers (or course they have our UUIDs now).

It doesn't look like we follow the very nice ordering that Adam did in PR #151, I think we're just full alphabetical (hopefully plus numeric too?) with no pinned concepts up front.

For the top-level dispatch, I think something like this would be a bit more defensive:
```ts
export interface WorkbookOptions {
  variant: ExportVariant;
  codingInstitution?: CodingInstitutionConfig;
}

export async function generateWorkbook(
  references: ReferenceSource,
  vocab: ConceptResolver,
  options: WorkbookOptions,
): Promise<XLSX.WorkBook> {
  switch (options.variant) {
    case "esea":
      return buildEducationWorkbook(references, vocab, options.codingInstitution);
    case "hpv":
      return buildHpvWorkbook(references, vocab);
  }
}
``` 
(Will have to update some existing tests to the approach.)

I think we're missing tests for:
- **Dropped-concept path** `buildHpvRows.ts:91-95` (record) + `:136-141` (`console.warn`):
  no test that a concept in no scheme / a scheme absent from `vocab.schemes` is omitted or warned. 
- **Header collision fallback** `buildHpvRows.ts:53-55`: no test for two same-labelled schemes.
- **`proxyBlobUrl`** `config.ts:52-59`: new exported fn, 3 branches, missing tests.

Applied concepts in *no* column are dropped with only `console.warn`. I don't think the real users of the spreadsheet will see this. (I also think it will be very rare).  Maybe an "Other codes" catch-all column?

danielbreves:
@jaybea thanks for the review!

We don't have access to EPPI ItemId in the bibliography. I added the OpenAlex ID, but it looks like none of the references have it.

I think I addressed your other comments!

jaybea (approved):
This looks great. So dramatically different (and better I think) than the ESEA spreadsheet. We really outputted the vocab url?

It fulfills all the requirements of Issue #127. My picky static code analysis issues thoroughly addressed, thank you.

Running it against staging, I was able to pull the json, and 1,217 of 1,231 of the refs do have EPPI ItemId's as `{ identifier_type: "other", other_identifier_name: "EPPI ItemId" }`. So we *could* pull them I believe. I'm totally open to whether we SHOULD output them or not. Just keeping in mind that **222** references have no DOI (and of course no OpenAlex ID) in the corpus.

If so, we could have an "EPPI ItemId" with one value per row, since "EPPI ItemId" is the only `other` kind in the coprus.

We might consider dropping OpenAlex ID, it has no entries in the entire full (HPV) corpus spreadsheet.

Nit: 
The `requestSearchExport` comment in `apiClient.ts` still says the export button is "disabled when q is empty". That gate was removed back when browse-mode export was enabled, so an empty search now exports the whole corpus (capped at 10k). Only the comment is stale, the behaviour is correct.

Image

danielbreves:
Thanks @jaybea. Good catch re "EPPI ItemId", I didn't realise it had been imported.

jack-o-the-green:
> We really outputted the vocab url?

For context on this, the vocabulary url exported in the esea spreadsheet is the vocabulary url from the linked data enhancement of each exported reference, which is different between linked data enhancements if they are coded with different versions of the taxonomy. It's a way to understand why certain columns for concept schemes might be blank (they may not have existed in previous taxonomy versions etc).

PR #174, opened 2026-06-17
Updates to filters (#171)

### LINKED ISSUE #171: Updates to filters
1. close the 'publication year' filter by default
2. move the 'Thematic focus - primary & secondary filters above publication year

danielbreves:
Closes #171.

- Collapse the **Publication year** filter by default for HPV.
- Move the **Thematic focus — Primary & Secondary** cards above Publication year (after Convening Theme).

Both are HPV-only. Implemented via a generic `defaultExpandedFilters` community config (parallel to `pinnedFilters`), which also folds the previously hard-coded year/country expansion into config. Absent ⇒ `["year", "country"]`, so Education is unchanged.

Screenshot 2026-06-17 at 4 44 01 PM

### DIFF
--- a/src/components/filters/FilterCardList.tsx
+++ b/src/components/filters/FilterCardList.tsx
@@ -5,7 +5,11 @@ import { YearRangeFilter } from "./YearRangeFilter";
 import { summary } from "./conceptSchemeFilterState";
 import { summary as countrySummary } from "./countryFilterState";
 import { summary as yearSummary } from "./yearRangeFilterState";
-import { orderFilterItems, type FilterItem } from "./filterOrder";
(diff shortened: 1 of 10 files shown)


jaybea (approved):
Well scoped and clean :)

FYI: Just noting that Issue #171 doesn't specify which of the 2 places the collapsed state should be done for, so we're assuming for both "Search" and "Visualise" and this PR does both, which makes sense to me.

One little test coverage nit to ponder noted inline.

jaybea:
Nitpick correction, only "year" is unit tested for the expansion settings, I implied "year" and "country" were in my nit.

PR #175, opened 2026-06-17
Implement popover tooltip

### LINKED ISSUE #170: Tooltips at the top of the filter drawer don't escape
Image

Needs to escape the header, and possibly be vertically shifted too.

Adam-Hammo:
Fixes #170 

Refactor tooltips to use [popover](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API) in order to properly escape above their own context.

Also, if a tooltip is too close to the top of the page, it will flip to be below the hovered element.

### DIFF
--- a/src/components/common/Tooltip.tsx
+++ b/src/components/common/Tooltip.tsx
@@ -41,7 +41,12 @@ function TooltipTrigger({
     const trigger = triggerRef.current;
     if (!shown || !bubble || !trigger) return;
 
-    bubble.showPopover?.(); // absent in jsdom and pre-2024 engines
+    // Re-runs when `text` changes mid-hover; showPopover() throws if already open.
+    try {
+      bubble.showPopover?.(); // absent in jsdom and pre-2024 engines
+    } catch {
+      // already open
+    }
 
     const r = trigger.getBoundingClientRect();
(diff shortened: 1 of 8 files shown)


src/components/common/Tooltip.tsx
```diff
@@ -14,9 +15,73 @@ interface TooltipProps {
...
+  // showPopover() lifts the bubble into the top layer, where it escapes every
+  // ancestor's overflow and stacking context (drawer/map scroll boxes) without
+  // a portal. We position it by hand for universal support; unmounting on hide
+  // clears the top layer, so no hidePopover() is needed.
+  useLayoutEffect(() => {
+    const bubble = bubbleRef.current;
+    const trigger = triggerRef.current;
+    if (!shown || !bubble || !trigger) return;
+
+    bubble.showPopover?.(); // absent in jsdom and pre-2024 engines
```
danielbreves: MDN says this throws `InvalidStateError` [DOMException](https://developer.mozilla.org/en-US/docs/Web/API/DOMException) if the popover is already showing. Should we have a guard here?

danielbreves (approved):
One minor comment, looks good otherwise!

PR #188, opened 2026-06-24
Add ai summary pdf download

### LINKED ISSUE #158: Allow a user to copy text from an AI summary
- when an ai summary is generated. allow a user to 'copy text' so they can save the summary elsewhere
- it should also allow the claims and sources
- wireframe idea below

Image

Adam-Hammo:
Closes #158 

Sample PDF: 
[ai-summary-cost-effectiveness-in-afghanistan-20260624-1-1.pdf](https://github.com/user-attachments/files/29276172/ai-summary-cost-effectiveness-in-afghanistan-20260624-1-1.pdf)


jaybea (changes requested):
First of all, this looks beautiful, nicely laid out header. Cut and paste in the doc works great. Nice to show the selected facets/search terms. Super convenient to have the search link clickable, really adds to the provenance.

Even quite readable with an unreasonable number of countries and facets (see image below).
Image

**Should fix** Some math-type symbols don't render in the PDF

I think we should fix the special character corruption in the resulting PDF. There's a lot of symbols in the these academic docs, and the mis-rendering even comes out like plausible looking wrong characters:  (10 μg dose) becomes 

Image

The jsPDF 4.2.1  standard fonts (Helvetica/Time/Courier) use WinAnsi encoding. A non-Latin-1 code point gets written as its 2 raw bytes and then reinterpreted as WinAnsi glyphs. 

This looks beautiful from the on screen summary: 

Image

But the PDF suffers:

Image

Maybe we could get a few of these in the standard ai summary so we can keep an eye on them.

**Nit (nice-to-have): filename date basis.**

Maybe we could use Locale for the filename instead of UTC? You'd get a funky date mismatch here in Australia for some times. If you switch it though take care about the unit tests now being dependent upon the timezone where the runner is, and some tests I think is in fixed in ..Z.

Adam-Hammo:
Great points @jaybea! I've addressed both. For the typesetting, the PDF generation now uses an embedded font which supports all the special glyphs.

jaybea (approved):
Excellent work, I love the comprehensive torture suite in the mock summarisation!

Issue #178, opened 2026-06-18
Investigation and export can handle both dicts and arrays for any object property field.


jack-o-the-green:
## Background
Our ESEA UI currently distinguishes between object properties it expects single values for (expects dictionaries, `documentType`, `ImplementerType`) etc and object properties it expects multiple values for (expects arrays, `educationTheme` etc). 

However, due to our current implementation https://covidence.atlassian.net/wiki/spaces/Platforms/pages/1242923013/2026-06-23+Vocabulary+Cardinality it is possible to receive either a dictionary or an array for a field we're expecting to be single valued. We want to update the UI to be more resilient to this. 

For example implementerType https://github.com/destiny-evidence/evidence-repo-ui/blob/685cde39b6c7d547838afbc64bcd1a16b6152668/src/services/investigationParser.ts#L192 is parsed out with `resolveCodedAnnotation` which expects a dictionary. However, it is not considered incorrect in the vocabulary to have multiple values in this field, and we may generate linked data enhancements that have multiple values here in future. Getting an array would make the parsing go kaput. 

The same is true for exports in most cases, for example study design. The UI currently expects a single study design for an export of the esea community https://github.com/destiny-evidence/evidence-repo-ui/blob/main/src/services/export/buildRows.ts#L348 and also it doesn't look like we're displaying study design on the results detail page and we probably should since it is possible to filter by. 

 
## Technical Scope
* Update the investigation parser in the UI to to handle both dicts and arrays for any object property field regardless of whether it has an `@set` container in the taxonomy context.jsonld
    * Don't forget the `parseAppliedConcepts` for HPV which is a special case where if a linked data enhancement has a single applied concept, we'll likely get a single URI here instead of an array https://github.com/destiny-evidence/evidence-repo-ui/blob/685cde39b6c7d547838afbc64bcd1a16b6152668/src/services/investigationParser.ts#L265
* Update the esea exports to handle both dicts and arrays for any object property field regardless of whether it has an `@set` container in the taxonomy context.jsonld 
    * It may make sense to more tightly integrate investigation parsing with exports so we only need to do this in one place.

jaybea:
implemented via #194 + #200

PR #194, opened 2026-06-30
export: tolerate dict-or-array JSON-LD shapes in the investigation-hierarchy export

jaybea:
## Summary

JSON-LD compaction emits a bare object for a single-valued property and an array for two or more, unless the property is declared `"@container": "@set"` in the vocabulary context, in which case it is always an array. Because the context controls this per property (and a new vocabulary version can flip it), any object-property field in the investigation hierarchy can arrive at the export as either shape.

This is the export half of #178 (the UI side of VMR#34). A companion change to the investigation parser and the detail-page display covers the read/display side (#200); #178 closes when both land.

## What changed

A small `ensureArray` / `first` pair normalises every object-property read in the row builders, applied across four resilience classes:

- **Coded concept columns** (e.g. `studyDesign`, `documentType`, implementer type, fidelity) emit all values, `; `-joined, via `ensureArray`.
- **Scalar columns** (intervention duration, sample size, attrition, cost): emit all values via `collapseCodedValues` (a lone value kept as-is with numerics preserved; two or more `; `-joined), rather than dropping the extras. Sample size, attrition and cost are gathered through the blank-node resolver (`resolveMany`) first.
- **Structural refs** (blank-node lookups, resolver, `refId`) tolerate a single-element-array wrapper.
- **JSON-LD containers**: repeatable ones (`hasFinding`, `hasArmData`, `hasEffectEstimate`) process every entry via `ensureArray(...).filter(isDict)`; singular ones (`hasOutcome`) take `first`. `hasFinding` is also normalised at the workbook entry point.

The coded concept columns join all values with `; ` so multiple codes are preserved; intervention duration likewise keeps all values (a single value kept as-is, multiple `; `-joined); the genuinely single-valued blocks (outcome, effect metric) take the first value. This is serialization resilience, not a new cardinality model: the export tolerates whichever shape each field arrives in, independent of how the data is serialized.

## Validation performed

- Typecheck clean (`tsc --noEmit`); 906 tests pass (`vitest run`), including new shape-tolerance and multi-value cases through the public `buildFindingRows` / `generateWorkbook` / `buildAllRows` contract.
- Real-data validation over the local ingest (ESEA + EEF only): ran the export over all 3,620 coded references loaded in a local repository instance (the complete local set, past the export endpoint's 10,000-row cap), through the real `buildAllRows` / `generateWorkbook` resolving against the ESEA v1.1 vocabulary. Against this data the `studyDesign` cell census reconciles exactly with the database: 3,620 rows built (none skipped), 4 with no study design, 3,162 single bare dict (one label), 454 array (58 genuine multi-distinct that join to distinct labels e.g. "Randomised Controlled Trial; Quasi-experimental study", plus 396 repeated-code arrays joined as-is). Zero `studyDesign` garbage cells; single dicts and arrays both materialise correctly through the SheetJS write path.
- Multi-value scan (FYI, scoped to the locally-ingested ESEA + EEF data; not a claim about production or future ingests): in this local set (3,620 references / 7,240 investigation enhancements) no record carries a multi-valued `duration`, `sampleSize`, `attrition`, or `cost`, and `hasOutcome` / `effectSizeMetric` are singular throughout; the fields that do carry multiples here (`educationTheme`, `studyDesign`, `setting`) are all on the join path. So within this data the duration change is a no-op; its value is forward-looking resilience for shapes the export could still receive.
- Earlier staging regression: a phonics search (21 records) exported through the full in-browser pipeline (server job, blob JSONL, client-side build); three sheets, switched concept columns resolve to prefLabels, no console errors.
- Pre-existing behaviour, not introduced here: 14 `ci_lower`/`ci_upper` cells render `NaN` because the source data carries the literal string `"NaN"` and `round5Cell` passes non-numeric values through unchanged. Byte-identical to `main` (same expression, same helper).

### DIFF
--- a/src/services/export/buildRows.ts
+++ b/src/services/export/buildRows.ts
@@ -206,6 +206,18 @@ function joinCodedValues(annotations: unknown): string {
 }
 
 /**
+ * Scalar `@value` cell: a lone value kept typed (numeric stays numeric),
+ * two or more `; `-joined so an extra coded value is never dropped.
+ */
+function collapseCodedValues(annotations: unknown): CellValue {
+  const values = ensureArray(annotations)
+    .map((a) => codedValue(a))
+    .filter((v): v is string | number | boolean => v != null);
+  if (values.length <= 1) return values[0] ?? null;
+  return values.map((v) => String(v)).join("; ");
+}
+
+/**
  * Concatenate the `supportingText` field of each annotation with ` | `
  * between entries.
  */
@@ -375,14 +387,15 @@ export function buildFindingRows(
   vocab: ConceptResolver,
 ): ArmRow[] {
   const resolve = makeResolver(buildBlankNodeLookup(findings));
+  const resolveMany = (value: unknown): PlainRecord[] => ensureArray(value).map(resolve);
   const rows: ArmRow[] = [];
   for (let i = 0; i < findings.length; i++) {
     const finding = findings[i]!;
     const armId = armIds[i]!;
     const ctx = resolve(finding["hasContext"]);
-    const sampleSize = resolve(finding["sampleSize"]);
-    const attrition = resolve(finding["attrition"]);
-    const cost = resolve(finding["cost"]);
+    const sampleSizes = resolveMany(finding["sampleSize"]);
+    const attritions = resolveMany(finding["attrition"]);
+    const costs = resolveMany(finding["cost"]);
     const intervention = resolve(finding["evaluates"]);
     const control = resolve(finding["comparedTo"]);
     rows.push({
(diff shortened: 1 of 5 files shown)


jack-o-the-green:
I'm being very pedantic about clarifications here, sorry in advance!

> JSON-LD compaction emits a bare object when a property has a single value and an array when it has two or more

This isn't quite right. This is correct _only_ if the `"@container": "@set"` is not present on the property in the `context.jsonld`. If the `"@container": "@set"` _is_ present on the property in the `context.jsonld`, we will get an array even when we get a single value. 

> so any object-property field in the investigation hierarchy can arrive at the export as either shape. 

This is still correct because a property without `"@container": "@set"` will arrive as either a dict or a list depending on the number of items. 

Because the vocabulary context controls the addition of that `"@container": "@set"` property, new vocabulary versions can flip whether a single item is a dictionary or a list. 

This makes me a iffy on whether the concept of scalar columns should exist at all in the UI - I can create a linked data enhancement with two intervention durations, and then export that reference from the UI, the UI is currently making the choice to drop my second piece of duration information. 

It would me most correct to remove that concept of scalar columns and keep all information in a linked data enhancement regardless. 

If it's a lot of extra work, I wouldn't bother changing it (because I don't feel like taxonomy changes are that frequent right now). But if it's easy to drop the `first` implementation in favor of joining multiple values in any column, lets do it.

jaybea:
Your @set correction is right, the PR body overstated it - I reworded it.

I made intervention duration and sampleSize/attrition/cost keep a single value as is type-wise, and build a joined string for two or more.

I kept `first` for outcome and effect metric since it seems ambiguous to have multiple values on those, and a second outcome would need its own rows rather than a joined cell I think. But I'm happy to adjust further if you think so.

Let me know!

jack-o-the-green:
> I kept first for outcome and effect metric since it seems ambiguous to have multiple values on those, and a second outcome would need its own rows rather than a joined cell I think. But I'm happy to adjust further if you think so.

Agree, good call! If hasOutcomes ends up being an array we have bigger problems.

jack-o-the-green (approved):
LGTM!

PR #195, opened 2026-06-30
Reference lists

### LINKED ISSUE #185: Download a bibliography/reference list
Whilst the excel export is great for users who want to be able to review - some users may want to share references with non-reviewer colleagues.

- we should add the ability to export a bibliography/reference list
- APA format 
- this should also be added to the bottom of the ai summary and it's pdf export

original request:
https://covidence.slack.com/archives/C07V94XKB29/p1782191727602339 
https://covidence.slack.com/archives/C0BC7AKAY7M/p1782381955440079

Adam-Hammo:
Closes #185 

- Adds RIS export plumbing
- Adds RIS->APA converter
- Adds export dropdown with added reference list PDF and `.ris` formats
- Adds reference list to AI summary drawer and PDF

Reference list caveats/notes:
- The author format varies per import source (eg HPV is "Last First" and EEF is "First Last") and it feels tenuous to try and unwind that here. APA prescribes `Last, F` but this formatter just passes through what the repository has.
- For the same reason, sorts by title instead of author last name. 
- Titles are meant to be in sentence case, but again we can't make our own distinction as to what's a proper noun or not so we preserve upstream title casing.
- By virtue of upstream data quality, references with OpenAlex bibliographies present better (due to having journals, published venues, etc). Deduplicating these over time will improve the amount of data available.

Sample files:
[ai-summary-cost-effectiveness-in-afghanistan-20260630.pdf](https://github.com/user-attachments/files/29491826/ai-summary-cost-effectiveness-in-afghanistan-20260630.pdf)
[evidence-repository-references-hpv-20260630.pdf](https://github.com/user-attachments/files/29491827/evidence-repository-references-hpv-20260630.pdf)
[evidence-repository-references-hpv-20260630.ris.txt](https://github.com/user-attachments/files/29491832/evidence-repository-references-hpv-20260630.ris.txt)
image


src/services/export/summaryPdf.ts
```diff
@@ -501,6 +465,28 @@ export async function buildSummaryPdf(
     gapAfter: 0,
   });
 
+  // ── References ────────────────────────────────────────────────────────
+  // APA bibliography for the summary's search, sourced from the backend RIS
+  // export (same as the results-page reference list). Plain text, alphabetical.
+  if (references.length > 0) {
```
jaybea, line 471: I think this could also be 0 if they're still loading, or if the load errored. I think you'd get a PDF w/o the offered references list.

Could be gated back to a disable on the download button (maybe allow it through on failed RIS since it's better than no PDF)?

jaybea (approved):
Obviously destiny-repository #767 needs to be deployed before this.

Tough call on the author ordering, yeah I think you made a good compromise.

Love pdfShared.ts pulling the DejaVu font embedding ino the shared module.

I imported into my fav biblio manager Paperpile, and authors, titles everything looks great. Except for type, they were all Misc (venut_type -> TY).

But I believe that needs to be addressed in destiny-repository.

Personally I'd love BibLaTex export, but agreed, it's a narrower audience and out of scope.

Only nit the race condition between the two jobs enabling the Download button (being honest, had to artifically slow the RIS report down) so your call whether to make the effort on that one. Although I guess a failed RIS report would be silently drop the section.

Adam-Hammo:
Thanks @jaybea, good points - let me think about the UX on the parallel work.

Re Misc types, this is unfortunately very common in the hpv/esea corpora as they've not yet been attached to the OpenAlex records (which have the richer bibliographic data for us to determine the type). I'd be open to considering changing the server-side default to JOUR instead of MISC? It'd be accurate most of the time but does have the potential to overpromise.

jaybea:
No, I'd keep the TY=GEN (as you have) until we get more authoritive typing sorted out.

Issue #60, opened 2026-05-20
Display Study Design in the UI


acdonald:
- Display study design to users if it has been coded. it should appear with the 'doc type' in reference details page
- Ensure that it is also displayed on the search results page
- When parsing the study design from the linked data enhancement, the parser should handle both a single concept (which will be a dictionary) or multiple study design concepts (which will be an array). 
- Exports should be updated to export multiple study designs if not already. 

<img width="706" height="718" alt="Image" src="https://github.com/user-attachments/assets/9a299aa7-c49d-4cae-88b5-9aec8c2a814f" />

PR #200, opened 2026-06-30
parser/display: tolerate dict-or-array shapes, preserve multi-values, and add study design

jaybea:
## Summary

For a property without `@container: @set`, JSON-LD compaction emits a bare object when it has a single value and an array when it has two or more, so an object-property field in the investigation hierarchy can arrive at the UI as either shape. Because the published context can change which properties carry `@set`, that shape is not stable across vocabulary versions either. The investigation parser and detail-page display previously assumed fixed shapes.

This is conceptually the parser + display half of #178 (the UI side of the vocabulary-mapping-robot change, VMR#34). The companion export change (#194) covers the write side. Because #194 is merged into this branch and has not yet landed on `main`, the GitHub diff for this PR also includes the export commits; they drop out of the diff once #194 merges. #178 closes when both land. Net-new UI: study design was filterable but never parsed or displayed, and now renders.

## What changed

The normalisation primitives `ensureArray` (take all) and `first` (take the first of any element type: dict, blank-node string, or bare CURIE) are hosted in `referenceUtils` and imported by the parser. Reads are made shape-tolerant across the same classes the export uses:

- **`referenceUtils`**: `getInvestigation` takes the first of an array-wrapped `hasInvestigation` (it was dict-only, so an array fell through to treating the enhancement root as the investigation); `extractFindingsAndEstimatesCount` counts a bare-dict single finding or estimate (it was array-only, rendering "0 findings / 0 estimates" for an unwrapped single).
- **Scalar and single-string annotation fields** (duration, sample size, attrition, cost, group differences, implementation name, funder, region) resolve to arrays and preserve every value rather than dropping to the first. The fields the export also emits (duration, sample size, attrition, cost, region) line up with its column join; group differences, implementation name, and funder are display-only. A field that arrives as a single object renders exactly as before.
- **Structural refs** (`evaluates` / `comparedTo` / `hasContext`, effect-estimate concept refs, arm `forCondition`) stay single via `first`, tolerating a single-element-array wrapper; the blank-node registry also registers array-wrapped definitions.
- **Applied concepts** (HPV `hasAppliedConcept`) tolerate a single bare URI or object, not just an array.
- **Coded concept fields go data-driven**: document type and study design (investigation), implementer type and implementation fidelity (intervention) are read as arrays and render every value as a tag. A single value looks exactly like before.

New and adjusted display:

- **Study design** renders as a tag group on the detail card and as result-row pills.
- **`InvestigationCard`** gates the metadata block with length checks so an empty array no longer emits an orphan divider, dedupes repeated concept codes by URI (multi-annotation data can repeat a code), and spaces stacked tag groups (the card only ever rendered one group before).
- **`InterventionDetails`** disambiguates multiple source-evidence entries by concept label (e.g. `Implementer: <label>`) instead of repeating a bare field label.
- **Multi-valued scalars** render as one labelled field with values joined: `; ` for finding-level scalars, `, ` for Region (matching the adjacent Country field).
- **Region source evidence** now appears in the context "Source evidence" toggle alongside Country; previously the region was displayed but its supporting text was dropped.

Cardinality is left to the taxonomy: the UI renders whatever shape and count arrive rather than hardcoding single-vs-multiple. An earlier revision kept the scalar and single-string fields single and deferred widening them; following code review that deferral is closed, so these fields now preserve all values. To keep the display in step with the export for the fields it emits, `sampleSize`, `attrition`, and `cost` also resolve blank-node string references (a definition carried on a sibling finding), and `attrition`/`cost` join the parser's pass-1 blank-node registry, matching the export's lookup. Genuinely single fields (structural refs, outcome, effect-size metric) still take `first`.

Out of scope: all upstream decisions about taxonomy / vocabulary-mapping-robot behaviour, including the `@container: @set` context fix, the `allowMultiple` annotation, and SHACL validation. This PR defends the UI against the shape inconsistency rather than fixing its source. Also deferred: value-qualified source-evidence labels. The field itself joins its values (Duration renders `2; 6`), but in the expandable Source evidence panel a scalar whose values each carry supporting text lists one entry per value under a repeated plain label (two `Duration` entries) rather than tagging each with its value (`Duration: 2` / `Duration: 6`). No record in the ingested corpus carries a multi-valued scalar, so this is latent in production data.

## Validation performed

- Typecheck clean (`tsc --noEmit`); 952 tests pass (`vitest run`), including multi-value shape-tolerance cases through the public `parseInvestigation` contract, attrition/cost blank-node resolution across findings, and component cases for multi-value field joins, region source evidence, tag rendering, the empty-array divider gate, dedup, and source-evidence disambiguation.
- Live-validated against a local repository instance through the dev server and the repository's API with the staging ESEA v1.1 vocabulary: injected multi-valued scalars (duration, sample size, funder, implementation name, region, group differences) each render joined into a single field; a second finding referencing another finding's `attrition`/`cost` blank nodes resolves and displays both values (parity with the export); the context "Source evidence" toggle surfaces the region supporting text. Console clean, no unresolved-reference warnings.
- Study design: the detail page renders the Study Design tag group; a record with 5 study-design annotations across 2 distinct codes collapses to 2 deduped tags; stacked Doc Type and Study Design groups are spaced correctly; search result pills surface both study designs (deduped, within the existing pill cap with a "+N more" overflow).
- Corpus check (direct database query): in the locally-ingested ESEA and EEF data, no coded record carries more than 2 distinct study designs (58 have two, the rest one), and the scalar fields are not observed multi-valued, so the widening is a no-op on current data and forward-looking for future ingests. Not a claim about production or other communities.
- HPV regression, live against staging (no injection): the `/hpv` search list renders applied-concept pills with the `+N more` overflow; a 90+-concept detail page renders the full Taxonomy Codes card (every scheme, hierarchical tags such as `Age Group › 9–14 years`, labels resolved from HPV vocab v2.3); console clean. The applied-concepts change only adds single-value tolerance on top of the existing array handling, so nothing HPV-specific changed.
- HPV reference-level Excel export, live against staging: the workbook still builds correctly (one References sheet, bibliographic columns plus one column per scheme, labels resolved, multi-values `; `-joined). The HPV export builder is untouched here; its only JSON-LD-shape reads (`getInvestigation`, `parseAppliedConcepts`) run through the same hardened functions as the display, so it inherits the resilience rather than duplicating it.

### DIFF
--- a/src/components/finding/InterventionDetails.tsx
+++ b/src/components/finding/InterventionDetails.tsx
@@ -42,13 +42,13 @@ function collectEvidence(intervention: InterventionData): SourceEvidenceEntry[] 
       "Implementation fidelity",
       intervention.implementationFidelities,
     ),
-    ...evidenceFrom("Implementation name", intervention.implementationName),
+    ...evidenceFrom("Implementation name", intervention.implementationNames),
     ...evidenceFrom(
       "Implementation description",
       intervention.implementationDescriptions,
     ),
-    ...evidenceFrom("Funder", intervention.funderIntervention),
-    ...evidenceFrom("Duration", intervention.duration),
+    ...evidenceFrom("Funder", intervention.funderInterventions),
+    ...evidenceFrom("Duration", intervention.durations),
   ];
 }
 
@@ -110,9 +110,9 @@ export function InterventionDetails({
         </div>
       )}
 
-      {intervention.implementationName && (
+      {intervention.implementationNames && intervention.implementationNames.length > 0 && (
         <LabeledField label="Implementation name">
-          {intervention.implementationName.value}
+          {intervention.implementationNames.map((n) => n.value).join("; ")}
         </LabeledField>
       )}
 
@@ -122,15 +122,15 @@ export function InterventionDetails({
         </LabeledField>
       ))}
 
-      {intervention.funderIntervention && (
+      {intervention.funderInterventions && intervention.funderInterventions.length > 0 && (
         <LabeledField label="Funder">
-          {intervention.funderIntervention.value}
+          {intervention.funderInterventions.map((f) => f.value).join("; ")}
         </LabeledField>
       )}
 
-      {intervention.duration && (
+      {intervention.durations && intervention.durations.length > 0 && (
         <LabeledField label="Duration">
-          {intervention.duration.value}
+          {intervention.durations.map((d) => d.value).join("; ")}
         </LabeledField>
       )}
 
--- a/src/components/finding/InterventionDetails.tsx
+++ b/src/components/finding/InterventionDetails.tsx
@@ -4,9 +4,13 @@ import {
   type SourceEvidenceEntry,
 } from "./SourceEvidenceToggle";
 import { LabeledField } from "../common/LabeledField";
-import { conceptsToTags, toHierarchicalTag } from "@/services/conceptLabels";
+import { conceptsToTags } from "@/services/conceptLabels";
 import { evidenceFrom } from "@/services/sourceEvidence";
-import type { InterventionData } from "@/types/investigation";
+import type {
+  CodedAnnotation,
+  InterventionData,
+  ResolvedConcept,
+} from "@/types/investigation";
 import "../common/LabeledField.css";
 
 interface InterventionDetailsProps {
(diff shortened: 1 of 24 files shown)


jaybea:
Displaying multiple values on the detail page from an EEF reference. Values injected for visualisation purposes.
image

A multiple study design output. Study designs are deduplicated. The paper below is from this reference (with the coding enhancements): https://data.evidence-repository.org/esea/references/doi:10.1037%2F0022-0663.96.3.403 

image

jaybea:
Some examples.

Multiple study designs now displayed in search results from https://data.evidence-repository.org/esea?q=%22Increasing+Reading+Comprehension+and+Engagement%22: 
image

...and in detail page:
image


Injected multiple values being displayed in previously single-valued cells on a detail page: 
image

jack-o-the-green:
Looks like this PR will close https://github.com/destiny-evidence/evidence-repo-ui/issues/60! Can you do a quick check to make sure all the scope is satisfied?

jaybea:
Yes, #60 scope fully satisified!

src/components/finding/InterventionDetails.tsx
```diff
@@ -16,26 +20,35 @@ interface InterventionDetailsProps {
...
   return [
-    ...themeEvidence(intervention),
-    ...evidenceFrom("Implementer", intervention.implementerType),
-    ...evidenceFrom("Implementation fidelity", intervention.implementationFidelity),
-    ...evidenceFrom("Implementation name", intervention.implementationName),
-    ...evidenceFrom("Implementation description", intervention.implementationDescriptions),
-    ...evidenceFrom("Funder", intervention.funderIntervention),
-    ...evidenceFrom("Duration", intervention.duration),
+    ...conceptEvidence("Theme", intervention.educationThemes),
+    ...conceptEvidence("Implementer", intervention.implementerTypes),
```
Adam-Hammo, line 40: Should we perhaps `dedupeByUri` here etc also?

jaybea: Yes, good point! I pushed the dedup down into the shared conceptsToTags so it covers all finding-detail surfaces, with InvestigationCard now sharing the same helper

